PR #50, opened 2017-12-12
clean up migrations

debboutr:


debboutr:
- remove 0002_auto_... | had been made w/ migrations removed
- removed migrations back to 0008
- makemigrations to rebuild all current models into 0009
- old migration 0010 was merged w/ 0002_auto

PR #95, opened 2018-01-19
Issue 61

debboutr:
this wasn't finished in the ticket that was accepted, but needed to be done.


dmlyons2:
@debboutr This one was set up to merge to `master` - did not catch it until after the merge :(

PR #100, opened 2018-01-20
Issue 79

debboutr:
couple of things with this ticket, I think I've got the functionality, but the interface may not be exactly what you want, I'm thinking that you may want a view of all products w/in a DataSource, instead this ticket just asks for a view of products w/in a DataSource that don't have a PUC assigned. This conflicts with a couple of things you had in the description

```3. Source category (if field Product.source_category is populated, view the Product.source_category, if Product.source_category is null this cell is blank in the table)```

**above this statement in the ticket you ask for this table to be a list of products w/o PUCs, so ALL of the cells are going o be blank in the table, and then below, similar issue, if they already have a PUC they aren't going to be in the table you are asking for anyway. I seems like you might want a view that looks at all products w/in a DataSource as well?**

```4. PUC (if a product category has not yet been assigned to the product, a button to 'Assign PUC' is enabled; if a product category has been assigned to the product user sees an indication that the PUC has already been assigned, e.g. text 'Complete')```

Have a look at what's here, and I can adjust before merging, if needed. Thanks
~rick


debboutr:
OH, I also need to mention that I've added DataSource as a foreign key in the DataDocument model, because of this I am re making the seed_data to reflect the change for testing

PR #111, opened 2018-01-27
Issue 103

debboutr:
**this ticket is not yet done, however, there are questions that I need answered before I can wrap it up.**

So far, I see 8 places where we are using tables. I have applied datatables functionality to the following:

- DataSource_list
- DataSource_view
- DataGroup_detail*

As I am looking into applying this functionality to the other 5 tables:

- ProductCuration_index
- LinkProduct_list
- CategoryAssignment
- QA
- ExtractionScript

In the datatables implemented so far, I have added DOM elements for all that is required in the ticket, i.e. pagination, search, no. records, download, and sort columns. In the DataGroup_detail view we have been using `<input>` checkboxes to display if the 'matched' or 'extracted' field is True or False. It doesn't seem like we want to have the functionality to change the value in these inputs AND if we want to download that data into the CSV, it can be done in the current style, but will take some hacking that probably isn't necessary, so that's the first question, can I change those `<input>` tags to just hold the string value from the model? we can paint the cell based on the value OR add an icon, etc.

As to the 5 other tables I haven't finished yet, it seems like having a CSV Download button may not make sense for all of them. So question 2 is which of these tables do we want to make downloadable? 

Hope this all makes sense. Talk soon


debboutr:
I have implemented datatables into all of the tables listed above, but for now only the top 3 tables have the 'Download CSV' capability, which can be added as needed.

PR #112, opened 2018-01-30
Dev webpack

zachmunzenrider:
I've started a page in the Wiki with explanation and instructions for configuring Yarn and Webpack in our dev environments


zachmunzenrider:
One thing I did NOT do with this push is remove the js libraries that have been added to our repo, which should go away once we have this running

Daveahern:
So @zachmunzenrider  I was able to run through all of your instructions, got all these new things installed, then got my local server running. How should I know whether this new JS package is included and running properly?

zachmunzenrider:
That's great Dave! The most basic thing is to view the source on a page and see that the aggregated js file is being served up, rather than the individual js libraries like jquery, datatables, etc.

zachmunzenrider:
Beyond that, a basic check that some js-dependent functionality in the system is still working properly. If you run into problems with this I can dig into it!

debboutr:
hmm, the branch is deleted now, but from what I saw this morning, all of the `<link>` tags were still in the `base.html`, so how do we know which resources are being utilized? because we still have the same static files that are being called in. OR were these libraries all rebuilt from the yarn command? _...puzzled_

PR #119, opened 2018-02-02
ZeroDivisionError bug fix w/ estimated records in DataSource

debboutr:
this was fixed in Zach's initial extraction script branch, but it never made it to `dev`, so I've added it here. We can't have zero estimated records if we want to know what percent of estimated records are uploaded or registered. I've taken the liberty to add the default value of 47 because it is the most common random number. In other words, 47 appears to be the quintessential random number of the universe.


Daveahern:
Merge pull request #119

PR #136, opened 2018-02-16
Dev 104

zachmunzenrider:
Search looks good!


Daveahern:
Ok so let's not merge this into master, it needs to be merged into dev.

Issue #107, opened 2018-01-24
quality metric for product categorization (8 points)


dmlyons2:
As Karyn the Curator
I want to track the level of curation for the assignment of a PUC to product
So that I can track and use the 'gold standard' PUC assigned to each product, and build confidence in overall data quality of PUC assignment to products.

A PUC can be assigned to a product in two ways:
1. Automatically (assigned by a machine learning algorithm) ('auto_puc')
  Application of the machine learning model to the data will be done *outside* the application. If a PUC is automatically assigned to a product, the following needs to occur:
  - record of which script assigned the PUC (script needs to have been registered in the extraction_script/scripts model, note see #108 for notes on how current extraction_script model will be generalized)
  - record date/time script was run
  - insert the assigned PUC into the Product.prod_cat field (**testing note, if a PUC has already been assigned to a product, that is if the Product.prod_cat field is not null, then there should NEVER be an automatically assigned PUC being inserted and overriding the PUC already populating the Product.prod_cat field).

2. Manually ('man_puc')
  Through the product curation functionality within our application, Karyn the curator can assign PUCs manually to products. If a PUC is manually assigned to a product in the product curation page of Factotum, the following needs to occur:
  - store the UserID of the person who did the manual selection/assignment of the PUC
  - store the date/time that the manual assignment of PUC was done
  - insert the manually assigned PUC into the Product.prod_cat field

PR #146, opened 2018-03-05
Ticket #107: metadata for PUC assignment

mwfrost:


debboutr:
Everything seems fine, couple of questions.

1. @ what point does the `Script` get associated?? or how? _looks like it's for a later ticket_
2. why add `#id` to the `<dd>` in the HTML template? _not being used and could later confuse organization of styling_

Also, it would be best to pull `dev` into this branch before merging, there will be conflicts w/ migrations, can be stemmed from 0025

nice work! we're cooking with gas now!

mwfrost:
Thanks, @debboutr . I've merged `dev` into this branch.

I added the #id selectors to the template so selenium can find the `<dd>` elements.

PR #147, opened 2018-03-07
37 dashboard stats

zachmunzenrider:
Adding document stats line chart to dashboard


templates/core/base.html
```diff
@@ -7,15 +7,31 @@
...
+        <link rel="stylesheet" href="{% static 'css/bootstrap.min.css' %}">
+        <!-- Open Iconic CSS -->
+        <link rel="stylesheet" href="{% static 'font/open-iconic-master/font/css/open-iconic-bootstrap.min.css' %}">
+        <!-- NVD3 CSS -->
+        <link rel="stylesheet" href="{% static "css/nv.d3.css" %}">
+        <!-- Custom CSS -->
+        <link rel="stylesheet" href="{% static "css/bootstrap_overrides.css" %}">
+    {% endblock %}
+
+    {% block javascript %}
```
dmlyons2: Not sure this is what you meant to do. This is a base template, so all of the other templates inherit this one. If you create a block in this template and put content in it, it will only show up if there is not a block javascript tag in the inherited template. 

Example if I make a dashboard template and inherit base, if I make a block javascript in that template it will replace what is in the base. Same is true for block css.

mwfrost:
the `date_format()` database function does not work with SQLite. `date()` is common to both MySQL and SQLite and should work.

PR #157, opened 2018-03-12
Issue 32 rd

debboutr:
you'll have to re-add the `seed_data` because the pk of the `ExtractedText` model is the same as the `DataDocument` model, it is a OneToOne relationship.

DataDocument <- 1:1 -> ExtractedText
ExtractedText <- 1:many -> ExtractedChemical

the `<div>` that holds the upload `<inputs>` will only appear if there are matched documents, but some or all not extracted. The idea I believe will be to extract all in one pass once everything in the csv validates.

there is a `test_extract.csv` in the `sample_files` folder that you can use to go with `Walmart MSDS 1` the other, `Walmart MSDS 2` has both ExtractedText records because they were already in the tests, so I kept them there.

The data that they have prepared will merge tables on the filename of the DataDocument, which will be given in the template when downloaded.

I have timed the loading of the 12 records in the `test_extract.csv` and am getting ~3.2 seconds, thisisn't quick, but in order to validate the way we need to we have to loop through the records twice.


mwfrost:
There is a cosmetic issue in how the upload form renders in Chrome.
sc_2018312_13635_194

PR #160, opened 2018-03-13
New seed data; correction to data source summary numbers

mwfrost:


debboutr:
I see yaml file with json in them, what am I looking at?

Daveahern:
@debboutr thanks for catching that, it looks like I forgot to put the yaml switch on my last dumpdata commands.

PR #186, opened 2018-03-22
149 dsstox_substance model

Daveahern:
Added model for dsstox_substance.
Removed non-model-creation methods from test_models.
Added in tests in test_models for creation of dsstox_substance object and related objects.
Merged origin/dev back into 149.


Daveahern:
Need to do this pull request for dev, not master.  Will re-create.

PR #200, opened 2018-03-30
Issue 162

debboutr:
this adds the **Download Registered CSV** button's functionality to DL a hidden table from w/in the page, we  may not need to have both this button as well as the one below **Export all records in {{ dg_name }} to csv**?? 

_I can add the info from that table into the first one before merging if that seems OK._


mwfrost:
Export all records in {{ dg_name }} to csv should probably be removed, but I'd prefer another product owner to confirm.

PR #219, opened 2018-04-24
Customize the test class used for functional testing

mwfrost:
This started as a 5-point ticket but grew in scope as I explored how to load the database fixtures only once, rather than with each test method. The new file, `functional_tests_rollback`, includes a custom class that inherits from `StaticLiveServerTestCase` but overrides the methods in `TransactionTestCase` that would otherwise cause the database to be reloaded and then deleted with every test method inside the class. Instead, this class mostly mimics the behavior of the `TestCase` class: it loads the database once when the Class is instantiated, then wraps each method in a transaction.


Daveahern:
If @zachmunzenrider or @debboutr can review this today, I'll get it merged in.

PR #226, opened 2018-04-30
datepicker refactoring

zachmunzenrider:


debboutr:
this looks good, but now that I think about it, we have the field `updated_at` in the form when we are creating a *new* datagroup, shouldn't this just default to `.now()`? And then, if we were to make changes to anything associated w/ the datagroup, i.e. DataDocument, ExtractedText, ExtractedChemical, should that change the `updated_at` field?

zachmunzenrider:
Rick, Kathie gave her blessing to removing the updated_at datepicker from the datagroup form, so I've done that, I'd recommend against expanding the concept of updated_at to include changes to relations with other models, though I agree this would be a useful thing to be able to see. (I had accidentally removed the datepicker in my other branch, but I've now removed it in this branch.)

PR #227, opened 2018-05-01
add inputs to link_product_form view and test

debboutr:
this adds 3 new input boxes to the view. the test asserts that the default `stub_xx` value loads into the page. However we may want to isolate the creation/linking of products <-> documents, we are currently searching for a product with the same title first and assigning that if it exists.


zachmunzenrider:
I made a small change to the view so that the app returns the user to the form if validation fails. Otherwise it looks good!

PR #245, opened 2018-05-04
ProductAttr model added for ProductCategory

debboutr:
this branch's migration will also likely need to be squashed and remade depending upon when it is brought into `dev`


zachmunzenrider:
Rick - I see you've changed the model name from `ProdAttr` to `ProdCatAttr`, which I guess makes it clearer, but the name is still really different from the naming convention we're using everywhere else. Why didn't you call it ProductCategoryAttribute? We spell everything else out

You're also including the `ProdCatAttr` class in the `ProductAttribute` model file, which isn't what we do elsewhere. Shouldn't that be in its own file?

Also, I think that model should be inheriting `CommonInfo`, like the other models do

debboutr:
these are all good points you raise. as far as the name, I feel anyone of us could look at `ProdCatAttr` and know what we are looking at, also it will help having to split calls across multiple lines if we need 11 chars instead of 24 chars. As far as where we put the `def` for the model it doesn't matter, we could have all of our models in one `models.py` file and it wouldn't matter, seeing as this model is only consumed by `ProductCategory` for now, I though it would be simplest to include in the same place. 
`CommonInfo` is already being inherited in the `ProductCategory` model and seeing as this is just a label I wouldn't think it important to know when the label was created, but if that were important we could certainly store the time if we wanted.

Daveahern:
@dmlyons2 we wanted to get your opinion on this question.  I think generally Matt and I agree with Zach's concern on standardizing on the convention of a separate file for each model and using a more complete name for the model, rather than having abbreviations in there.

I think we've all agreed on our standup call that inheriting from CommonInfo isn't necessary for lookup tables, since we won't need to track creates/updates of lookup lists.

Do you have strong thoughts about these model conventions?

dmlyons2:
I concur with @zachmunzenrider on on standardizing on the convention of a separate file for each model and using a more complete name for the model, rather than having abbreviations in there.

Daveahern:
@zachmunzenrider  if you can do a quick review to make sure your concerns are addressed, I can try to merge this in before our review.

PR #269, opened 2018-05-21
Dev 123

zachmunzenrider:


debboutr:
do we always need to build the intermediary join table on `ManyToMany` relationships? the docs state that Django will do this. I'm not very experienced in DB's, but there are going to be *3* `ManyToMany` Fields on the `Product` model so it might be better to figure out these relationships sooner than later. It seems to me the only reason for creating the `through` intermediary table is to have the ability to store dates? Is that something that is important when linking `ProductAttributes` to `Product`? We may be able to get by with less if not? `https://docs.djangoproject.com/en/1.11/ref/models/fields/#manytomanyfield`

mwfrost:
I think adding the mediating table makes sense if we might put attributes on the join.

Daveahern:
Yeah, for example, the Product to PUC table is required for that many to many because of the other fields like classification_method etc that get stored there.

zachmunzenrider:
Zach [2:49 PM]
It does seem to me like a small extra effort to model the mediating table explicitly, for the benefits Matt and Dave describe. Even if we don't have immediate plans to add attributes, it gives us flexibility

@rick I wasn't sure of a better test to write for a model test, but I'd love to see what you do!

PR #270, opened 2018-05-23
232 product puc linker

Daveahern:


dashboard/models/product.py
```diff
@@ -2,50 +2,56 @@
...
-		ordering = ['-created_at']
\ No newline at end of file
+    def get_uber_product_to_puc(self):
+        pucs = self.producttopuc_set
+        if pucs.filter(classification_method='MA').count() == 1:
+            return pucs.filter(classification_method='MA').first()
+        elif pucs.filter(classification_method='AU').count() == 1:
+            return pucs.filter(classification_method='AU').first()
+        else:
+            return "You're crazy to think this product has been categorized!"
```
mwfrost: Nice, but it breaks. The subsequent `get_uber_puc()` method is expecting `get_uber_product_to_puc()` to return a `PUC` object, not a string. Can you try returning a `None`?

Daveahern: Yeah good point.  I think since get_uber_puc() expects a PUC it will barf on a None so I'll have to handle that case there as well.

Daveahern: This works better
```
 def get_uber_product_to_puc(self):
        pucs = self.producttopuc_set
        if pucs.filter(classification_method='MA').count() == 1:
            return pucs.filter(classification_method='MA').first()
        elif pucs.filter(classification_method='AU').count() == 1:
            return pucs.filter(classification_method='AU').first()
        else:
            return None

    def get_uber_puc(self):
        thispuc = self.get_uber_product_to_puc()
        if thispuc is not None:
            return self.get_uber_product_to_puc().PUC
        else:
            return None
```

Daveahern:
Ok so this has been updated now to throw an error when a second manual PUC assignment is attempted.  The app doesn't allow for it in the normal workflow, so after doing a first PUC assignment to a product, return to the /product_puc/<pk> page for the one you just assigned and try to do it again.  It should have an error raised. Verify in the DB that a second manual PUC assignment has not been saved to the producttopuc table.

PR #275, opened 2018-05-30
Dev 123 - product attribute models

zachmunzenrider:


dashboard/migrations/0042_auto_20180521_1705.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
```
Daveahern: I think there's a better way to do this migration.  Since migrations were created simultaneously on two branches, we now have two that are numbered `0042`.  It works, but I think the numerical sequencing is confusing when we'll talk about them.  It seems that the better thing to do would be to go back to the way your database was on `dev` - Blow your database away and while on the `dev` branch migrate up to `0044`.  Then switch to branch `123` and delete your new `0042` and `0045` migration files.  Then run makemigrations again.  This would create a single new `0045` migration to create ProductAttribute.  What do y'all think @zachmunzenrider @debboutr ?

zachmunzenrider:
Hey @Daveahern can you review my changes? I think these address your concern.

Daveahern:
Everything looks good.  The new migration simplifies things.  Admin interface to add product attributes works and the new test files pass.

PR #277, opened 2018-06-04
Issue 126 revised

dmlyons2:
**Note that a migration is required.**  Changed @debboutr original code to make the PUCAttribute not required. If it is required it will not migrate to `prod` as there are already 100's of PUCs. Instead of putting in a false value for existing data I made it a non required field.    

No test added for this ticket.


dashboard/models/PUC.py
```diff
@@ -11,6 +11,7 @@ class PUC(CommonInfo):
     last_edited_by = models.ForeignKey('auth.User', default = 1, on_delete=models.CASCADE)
     products = models.ManyToManyField(Product, through='ProductToPUC')
 
+
     class Meta:
         ordering = ['gen_cat', 'prod_fam', 'prod_type']
         verbose_name_plural = 'Product categories'
```
mwfrost: This should say "Product use categories"

PR #282, opened 2018-06-15
239 source group doc changes

Daveahern:


debboutr:
I don't understand why there needs to be a ForeignKey relationship from `DocumentType` to`GroupType`? Doesn't this just parallel the same logic that we have from `DataDocument` to `DataGroup`? 

Alternatively, couldn't these 2 new `Type` models be reduced into 1 that could then hold a field let's call it `source` for now that would hold a value of either `DG` or `DD` and then in the `DataGroup` and `DataDocument` models we could limit the choices like so:

`document_type = models.ForeignKey(SourceType, on_delete=models.PROTECT, null=True, 
blank=True, limit_choices_to={'source': 'DD'})`

this assumes the model would again be called `SourceType` with the fields [`title`,`description`,`source`]

could be done similarly in the `DataGroup` model with: `limit_choices_to={'source': 'DD'}`

If nothing else, I question the need to build another hierarchy w/in types that we already have w/in the models that will consume them w/ the foreignkey relationship between the 2 new type models.

Daveahern:
Actually Rick in your proposal, how would you restrict which DocumentType can be chosen given a GroupType? 

So given this data:

Data group type 1 = composition
Data document types = MSDS, SDS, or ingredient disclosure

Data group type 2 = Habits and Practices
Data document types = journal article, professional society report, or government report

If data documents are being uploaded for a data group type 1 (composition), then the system should be able to enforce the choice that the only data document types that can be assigned are in the child DD type list (MSDS, SDS, ingredient disclosure) and the system knows how to enforce this because of the FK in the DocumentType record that points to its parent.

I'm not sure how this would be done in your suggestion, other than by adding an additional "parent" ID in your combined type table, which would essentially be the same FK we're discussing now, it's just self-referential to the same table. And that structure would certainly work, and is something we did talk a bit about in person in RTP, though I think in reference to the PUC hierarchy.

PR #290, opened 2018-06-28
creating dashboard and qa tests

zachmunzenrider:
Note that I'm using lxml as a lightweight alternative to Selenium in these tests. Everything that Selenium was being used for in the older version was HTML scraping. I added it to my environment using Anaconda. (When I tried to install it with pip I got some dependency errors; YMMV)


Daveahern:
@mwfrost @debboutr can one of you try to review this?  I've been unable to get the `lxml` library to work, it gives me errors when it calls `etree`.  Rather than fight with it, maybe it will work well for one of you.

PR #292, opened 2018-06-30
273 add org to data document

zachmunzenrider:
I've added a temporary fixture to aid in testing this ticket. After doing a migration to add "organization" to the data_document model, run a loaddata on temp_habits_and_practices.yaml.

Once that's done, you should be able to create a new datagroup of type "Habits and Practices", using the habits_and_practices_sample_registered_records.csv from sample_files. If all goes well you should see 4 new datadocument records with organization values.

NB: The datagroup update form does NOT do anything with data_documents when you specify a new registered records csv file. I wanted to check in with folks that this is ok, because it did confuse me during development.


Daveahern:
The code changes make sense and the functionality works when I test it.  I think we may have had a discussion in slack, but as written the organization is not required for H&P group types (this matches what is specified in the issue) but it is also not forbidden for other group types.  The issue is vague on this point, so I consider that reasonable.

One suggestion would be to add a test @zachmunzenrider .

PR #294, opened 2018-07-02
244 upload extracted text for habits and practices

zachmunzenrider:
Should be pretty straightforward to test: just models and a model test. . . .


debboutr:
@Daveahern `null=False` does not influence whether or not the field is required or not, read the docs
`https://docs.djangoproject.com/en/1.11/ref/models/fields/#null`

PR #297, opened 2018-07-03
Issue 46 rd

debboutr:
the test made for this ticket is not that great, but it is useful. I've just noticed that when all objects are approved and the user is sent back to the `qa` page, the button still says `Continue QA` even though there is nothing left to QA. There is a whole lot of refactoring that could be done after all of this, the way we load the forms, how we come up with the "next" text record and then implement functionality, and how elements are enabled/disabled. If there is an error with an edit, the page is loaded with all of the form elements disabled and you need to toggle again with the button. Anyway, I think everyone should take a look at this and maybe we could all make a list of things that we would want to see improved?


Daveahern:
So I've been banging on the functionality here and it's working very well.  I've worked though various scenarios including making edits but not approving a record then coming back in later to make more edits and approve.  One oddity in the interface I've noticed is the difference in the "QA Checked" button across browsers and in fact a bad icon on Firefox. It's minor and I don't think affects the review here.
qa_checked_difference

debboutr:
one thing that I think we'll want to add here is changing the `qa_begun` attribute of the `Script` model once all of the "extracted" objects are approved. This way when you are returned to the `/qa/` view you will then be able to have the option to "Begin QA" again on that same `Script` object?  or would we want to add to that view a query like so:

```len(ExtractedText.objects.filter(extraction_script=<sc>, qa_checked=True)) == len(ExtractedText.objects.filter(extraction_script=<sc>))```

this isn't the best way to do it, but it illustrates the idea. If I remember correctly, we will still want to QA an extraction script once it has already had a single QAGroup created and then subsequently approved, i.e. we will always want to see ALL of the extraction scripts listed in the `/qa/` view, whether or not QA is complete on any QAGroup w/in them.

PR #302, opened 2018-07-10
fix namespaces so all tests run, fix test_extracted_text_qa_notes test

debboutr:
I fixed the qa_notes test that matt had added but wasn't being run, so the fail never showed and changed the class names so that all tests will be run, we should currently have 27 tests at this point.


Daveahern:
@dmlyons2 Me too.  I've been sitting here on my EPA laptop just staring at the progress bar.

PR #310, opened 2018-07-17
Dev patch ztm

zachmunzenrider:


dashboard/migrations/0057_auto_20180717_1331.py
```diff
@@ -0,0 +1,35 @@
...
+        ),
+        migrations.AlterField(
+            model_name='qagroup',
+            name='extraction_script',
+            field=models.ForeignKey(limit_choices_to={'script_type': 'EX'}, on_delete=django.db.models.deletion.CASCADE, to='dashboard.Script'),
+        ),
+        migrations.AlterField(
+            model_name='taxonomy',
+            name='last_edited_by',
+            field=models.ForeignKey(default=1, on_delete=django.db.models.deletion.SET_DEFAULT, to=settings.AUTH_USER_MODEL),
```
Daveahern: If a user is ever deleted, these records won't be deleted, but the user attribute will be set to the default user (id=1).

dashboard/models/taxonomy.py
```diff
@@ -3,19 +3,16 @@
...
-
-	class Meta:
-		verbose_name_plural = 'Taxonomies'
-
-	def __str__(self):
-		return str(self.title)
-
+    title = models.CharField(max_length=100, blank=False, null=False)
+    description = models.TextField(null=True, blank=True)
+    parent = models.ForeignKey('Taxonomy', null=True, blank=True, on_delete=models.CASCADE)
```
Daveahern: Django 2.0 requires these on_deletes to be explicit now.

dashboard/search_indexes.py
```diff
@@ -7,7 +7,7 @@ class ExtractedChemicalIndex(indexes.SearchIndex, indexes.Indexable):
     text = indexes.EdgeNgramField(
         document=True, use_template=True,
         template_name = 'search/indexes/dashboard/extractedchemical_text.txt')
-    title=indexes.EdgeNgramField(model_attr='raw_chem_name')
+    title=indexes.EdgeNgramField(model_attr='raw_chem_name', null=True)
```
Daveahern: This is an unrelated bug that was causing the indexing to fail if a raw_chem_name had a null value in it.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 certifi
-Django>=2.0
+Django>=2.0.7,<2.1.0
```
Daveahern: Django 2.1 is too far, so we insist on 2.0.

PR #320, opened 2018-07-18
Exclude DSSToxSubstance and ExtractedChemical from the search results

mwfrost:
The `/find` page now explicitly filters on results whose `facet_model_name` is either "Data Document" or "Product". 

I also added a few new files: before I excluded those models from the search results page, it was throwing errors related to the lack of a view to handle the URLs for DSSTox Substance objects.


mwfrost:
The real fix is in `dashboard/views/search_forms.py`.

PR #323, opened 2018-07-18
change css selector for QA table to match others

mwfrost:


Daveahern:
@mwfrost what about the `extraction_script.html` template?  It's also using `thead-inverse` and not `table-primary`

mwfrost:
@Daveahern Maybe we need a ticket for "review css selectors for all tables"?

PR #336, opened 2018-07-24
258 restructure test directories

dmlyons2:


dashboard/tests/test_extracted_qa.py
```diff
@@ -1,25 +0,0 @@
-from django.test import TestCase
```
Daveahern: Did you determine these tests were no longer needed?  I'm not sure where else we are testing these qa processes.

dashboard/tests/test_qa_approval.py
```diff
@@ -1,47 +0,0 @@
-from django.test import Client
```
Daveahern: @dmlyons2 Did you determine these tests were no longer needed?  I'm not sure where else we are testing these qa processes.

dashboard/tests/test_extracted_qa.py
Daveahern: Hmm, were these not being run?  I don't see them in the original init file that's been changed here.

PR #348, opened 2018-08-01
Chemical summary for Get Data page

mwfrost:
This is a much larger PR than the ticket would suggest, because it includes a lot of changes to the seed data that are unrelated to the content of the ticket.

The new seed data suite should be used for all subsequent dev work, so we have a common starting point when we make changes.


debboutr:
can you add an example CSV into samplefiles for uploading? I just tried grabbing the first 6 DTXSIDs and created a csv, but I just get an empty output w/ these headers:

DTXSID | pucs_n | dds_n | dds_wf_n | products_n
-- | -- | -- | -- | --

here are the values that I used..

```
In [2]: a = DSSToxSubstance.objects.all()[:6]

In [3]: a.values_list('sid',flat=True)
Out[3]: <QuerySet ['DTXSID9022528', 'DTXSID7020182', 'DTXSID6026296', 'DTXSID2021781', 'DTXSID9022528', 'DTXSID9022528']>
```

debboutr:
in `get_data.py`
```
    stats = pucs_n\
    .annotate(dds_n=Value(-1, output_field=IntegerField())) \
    .annotate(dds_wf_n=Value(-1, output_field=IntegerField())) \
    .annotate(products_n=Value(-1, output_field=IntegerField()))
# could be written as...
    v = Value(-1, output_field=IntegerField())
    stats = pucs_n.annotate(dds_n=v).annotate(dds_wf_n=v).annotate(products_n=v)
```
Also,
```
        for line in lines:
            if DSSToxSubstance.objects.filter(sid=str.strip(line)).count() > 0:
                dtxsids.append(str.strip(line))

# this seems easier to read and doesn't duplicate stripping

        for line in lines:
            val = str.strip(line)
            if DSSToxSubstance.objects.filter(sid=val).exists():
                dtxsids.append(val)
```
you might spend some time to try to make the `strsql` object more readable to others as well using parens and making a multi-line string:

```
 ("SELECT dss.sid , "
    "IFNULL(SUM((SELECT Count(DISTINCT ec2.extracted_text_id) as dd_wf_id "
        "FROM dashboard_extractedchemical ec2 "
        "WHERE ec2.id = dss.extracted_chemical_id "
        "GROUP BY ec2.extracted_text_id "
        "HAVING SUM(("
            "(ec2.raw_max_comp IS NULL)+"
            "(ec2.raw_min_comp IS NULL)+"
            "(ec2.raw_central_comp IS NULL)) = 0)"" > 0 )),0) as dds_wf_n "
    "FROM dashboard_dsstoxsubstance dss "
    "LEFT JOIN dashboard_extractedchemical ec on ec.id = dss.extracted_chemical_id "
    "GROUP BY dss.sid ")```

Daveahern:
It works well from a user interface perspective.  One aspect that I was confused about was what to upload.  Without instructions or a template, I took a guess and uploaded a csv file with a single column of DTXSID's.  It worked!  The issue says a template should be available to download but I don't see one.  It's a bit odd, since a csv template for a single column of data is pretty minimal. I could see some instructions for filling out the template would be worthwhile.

PR #360, opened 2018-08-05
move bulk button, add product column w/ link

debboutr:
currently the button that is in the 'Product' column will be filled with the title of the product, which if bulk-created will end up being 'unknown' for potentially many of the rows in this table. Probably not a big deal, but worth mentioning.

### DIFF
--- a/dashboard/tests/functional/test_datagroup_detail.py
+++ b/dashboard/tests/functional/test_datagroup_detail.py
@@ -135,3 +135,15 @@ class DataGroupTest(TestCase):
         span = '<span class="oi oi-circle-check" style="color:green;"></span>'
         self.assertIn(span, response,
                       'Check should be present if matched.')
+
+    def test_detail_table_headers(self):
+        pk = self.objects.dg.pk
+        response = self.client.get(f'/datagroup/{pk}').content.decode('utf8')
+        self.assertIn('<th>Product</th>', response,
+                      'Data Group should have Product column.')
+        fu = GroupType.objects.create(title='Functional use')
+        self.objects.dg.group_type = fu
+        self.objects.dg.save()
+        response = self.client.get(f'/datagroup/{pk}').content.decode('utf8')
+        self.assertNotIn('<th>Product</th>', response,
+                      'Data Group should have Product column.')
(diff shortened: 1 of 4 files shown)


dashboard/tests/functional/test_datagroup_detail.py
```diff
@@ -135,3 +135,15 @@ def test_delete_doc_button(self):
         span = '<span class="oi oi-circle-check" style="color:green;"></span>'
         self.assertIn(span, response,
                       'Check should be present if matched.')
+
+    def test_detail_table_headers(self):
+        pk = self.objects.dg.pk
+        response = self.client.get(f'/datagroup/{pk}').content.decode('utf8')
+        self.assertIn('<th>Product</th>', response,
+                      'Data Group should have Product column.')
+        fu = GroupType.objects.create(title='Functional use')
```
dmlyons2: `fu` as a variable name. Nice

Daveahern:
What happens where there is a negative number of products to create?
negative_bulk

debboutr:
I just added a line in the `link_product_form` view that will prevent there from being more than one link from DataDocument to Product, but as I did this I realized that if the user were to navigate back to this view with the url to that specific DataDocument, they would then potentially enter data, hit submit and then never know that there is already a link and that their entered data is getting thrown out, except for `document_type` we might want to render something that says the document has already been linked if this is the case?

dashboard/views/product_curation.py
```diff
@@ -111,8 +113,9 @@ def link_product_form(request, pk, template_name=('product_curation/'
                 product.size = form['size'].value()
                 product.color = form['color'].value()
                 product.save()
-            p = ProductDocument(product=product, document=doc)
-            p.save()
+            if not ProductDocument.objects.filter(document=doc).exists():
```
Daveahern: This looks good. So if there's already a Product for a DataDocument it won't create a new one.  So we'll never have more than one - fixing the issue with the negative numbers showing up on the create product button.

Daveahern:
@debboutr I think we don't need to worry about a user manually entering in an old URL.  There are other cases too where they might do this and the behavior would be unexpected.

PR #375, opened 2018-08-10
View Habits n Practices in get_data view

debboutr:


zachmunzenrider:
A few things:
1. I can get now search for matching PUCs on the Get Data page, but without any Habits and Practices seed data I'm unable to review the resulting cards. Rick, did you create some seed data you could add to the fixtures?
2. The test with this ticket seems like just a stub--it isn't doing the test described in the ticket.
3. I fixed a bug that was blocking me in get_data.py and pushed it up, along with merging the changes to dev, which didn't seem to cause a conflict.

mwfrost:
The `puc_autocomplete.py` file was specifically coded to return results only to logged-in users. The Get Data page is meant to be accessible to unauthenticated viewers. One of those things needs to change before this works properly for a user who is not logged in.

mwfrost:
There is now a good test of the content that is returned by a search, the PUCs are available to unauthenticated users, and all the tests pass.

The formatting of the numeric values still needs to be fixed to suit the requirements in the ticket:
```
mass along with mass_unit (e.g. if mass = 4, mass_unit = gm, user sees 'mass: 4 gm')
```
The current version expresses the mass values with trailing zeros.

debboutr:
@kdionisio how do we want to see the `mass,frequency, and duration` fileds in the output of the card? We have made them `DecimalFields` so I would think that there might be a need to show the decimals. Is there a particular number of them that you want to see? @zachmunzenrider had specified 5 decimal places in the model for these fields so I'm imagining that we want them, but I don't read into the ticket that there should be a particular number of places shown or that non of them should be shown, just that you want the number placed next to the unit description. So, if we enter `4` as a value for mass right now the output will be `4.00000` because we specified in the model to hold 5 decimal places. Let me know if you would rather see another number of decimal places or none at all.

kdionisio:
@debboutr My apologies as I was not specific on this point in the ticket. We certainly will have entries where decimal places are included, and where we want to see the decimal places. For simplicity, why don't we always view 2 decimal places on the application (so may show as 4.00, or as 0.05).

Issue #371, opened 2018-08-07
error when uploading extracted text (5 pts)


kdionisio:
When trying to upload extracted text (http://factotum.epa.gov/datagroup/62) to a data group with group type 'composition,' I receive an error when the completed extracted text .csv is uploaded (see below). Error seems to be with the 'raw_cas' column header, but this is the column header provided in the extracted text template which is downloaded.

Request Method: | POST
-- | --
http://factotum.epa.gov/datagroup/62
2.0.6
TypeError
'raw_cas' is an invalid keyword argument for this function
/home/dlyons/anaconda3/envs/factotum/lib/python3.6/site-packages/django/db/models/base.py in __init__, line 495
/home/dlyons/anaconda3/envs/factotum/bin/python
3.6.4
['/data/code/factotum',  '/home/dlyons/anaconda3/envs/factotum/lib/python36.zip',  '/home/dlyons/anaconda3/envs/factotum/lib/python3.6',  '/home/dlyons/anaconda3/envs/factotum/lib/python3.6/lib-dynload',  '/home/dlyons/anaconda3/envs/factotum/lib/python3.6/site-packages']
Tue, 7 Aug 2018 16:32:07 -0400

kdionisio:
Unilever data documents

Other information you may need if starting from scratch:
Group type: Composition (need to select this when creating the data group)
Weight fraction type: reported (need to select this when uploading the extracted text file)
Extraction script: required but can choose anything (need to select this when uploading the extracted text file)

[1.pdf](https://github.com/HumanExposure/factotum/files/2271517/1.pdf)
[2.pdf](https://github.com/HumanExposure/factotum/files/2271518/2.pdf)
[3.pdf](https://github.com/HumanExposure/factotum/files/2271519/3.pdf)

PR #378, opened 2018-08-13
371 fix extracted text upload error on dev

mwfrost:
This branch applies the same fix to `dev` that the 371 hotfix branch applied to `master`. It also includes a functional test that generates a csv file containing ExtractedChemical records and uploads it through the HttpRequest object.


mwfrost:
Oh yea, right: that `self.in_mem_csv = in_mem_sample_csv` assignment was carried over from some sample code that I was afraid to deviate from.

debboutr:
issue #371

PR #394, opened 2018-08-29
362 abstract filenames

debboutr:
this will satisfy the filenaming convention moving forward, but we will still need to rename files that have already been uploaded.


mwfrost:
I even uploaded a registered record with a file name that contained two periods, and it seemed to work.

PR #402, opened 2018-09-04
363 datagroup directories

Daveahern:
Data group files will now be stored in a directory .../media/<pk>/.  Tests for DG media have been updated and pass.  I have a draft script that can be used to rename existing DG directories and update the csv and zip fields in DG records that I'll post separately and @dmlyons2 can review/update/use for deploying this change to production.


Daveahern:
Err, I merged origin/dev back into this branch earlier today so I'm not sure why there are conflicts.  Maybe I'll go do that again before someone looks at this.....

mwfrost:
From above:

> I don't see a reason to keep the dgurl method on the DataGroup model if all it is doing is returning the pk. grep for the instances of dgurl in the dashboard and replace w/ the pk, then you can delete the method in the model class.

Here's precisely why: we might need to change the way we assign and derive the media folder path for a data group object. It's a drag to hunt through the code and evaluate every case where the data group's `id` is read. It would have been better to keep that method in place.

I have committed some changes that take care of the cases I've spotted so far, but our test coverage has been shown to be insufficient. For example, I tried to download a data group's registered records as a csv using the link on the data group detail page, but it failed because the `+` operator had been used to concatenate strings. There was no test to pick that up -- I plan to write one.

mwfrost:
I just pushed 47 commits to this branch, because of all the merging into dev that just happened. The commits distinct to this ticket pertain to the data group's media folder and the way the application builds paths to it.

mwfrost:
I just pushed my new test changes.

debboutr:
looking at this now, we had put the default name in the form like so:
```
    group_key = DataGroup.objects.filter(data_source=datasource).count() + 1
    default_name = '{} {}'.format(datasource.title, group_key)
```
which will mean that if we look for the datagroup on just name alone:
```
    def get_name_as_slug(self):
        return self.name.replace(' ', '_')
```
it won't get us to the default, nor if the user had changed the name in the form before saving will it get us to where that directory is. This is probably a good argument to go through and rename all directories w/ a uuid and update the `fs_id` field. We should be able to `split()` the `zip_file` string to get the directory though.

couple other small things. I don't know that we would really want to use the full path in these string by adding `settings.MEDIA_ROOT`. I don't think that gains us anything. Also, when using f-strings in python3 you don't need to `format` objects like is done here:
```
uuid_dir = f'{settings.MEDIA_ROOT}{str(self.fs_id)}'
uuid_path = f'{self.get_dg_folder()}/{str(self.fs_id)}.zip'
zip_file_path = f'{self.get_dg_folder()}/{str(self.get_name_as_slug())}.zip'
```
none of these need the `str()` function.

I just dropped my schema and did a `loaddata` w/ what `363_...` has for seed data and it seems the UUID's are all automatically created for each DataGroup even though it isn't in the seed data, so these are going to change everytime we wipe our `dev` DB's. this isn't likely a problem, but something to be aware of.

Issue #164, opened 2018-03-15
format faceted search results (5 pts)


kdionisio:
As Karyn the curator
I want to view faceted search results in a tabular layout with basic search/sort capabilities
So that I can more easily absorb information from the search results, and organize search results for further analysis.

Please change layout of faceted search results to be in a table format with basic functionality (pagination,
 search within table, etc.) (e.g., implement data tables, or similar)

mwfrost:
The faceted search returns heterogeneous objects, and the table can only show attributes common to all the objects. That should not be a problem, since the attributes we are returning with search results can be stacked and shown together.

kdionisio:
**edited to note searching/sorting capability within table is not required at this time.

PR #403, opened 2018-09-06
164 format search results

debboutr:
#164 

not sure if the test is the best for this ticket.

There is a `product_to_puc` seed data yaml that you can load, then rebuild_index, and if you search "terro" you will see some pucs in that column.

We should probably add a color legend if we decide to keep colors in the output. Also, I'm only guessing at what appropriate column names for the table would be


zachmunzenrider:
Running into an issue on your ticket where the sorting and searching is only working on the current page when there are multiple pages. E.g., if I sort by Type, it orders the visible records on the page by type, but I think it should be ordering the whole table, not just the current page. Likewise, when I search using the search box, it only finds matches on the current page.

I think the workaround is to have DataTables handle everything client-side, rather than relying on the server for paging. I know in my other project I have some quite large tables handled this way and it seems to work well. Does anyone else have thoughts about the best way to handle this?

Issue #311, opened 2018-07-17
Remove uploaded_at field from DataDocument model (5 pts)


mwfrost:
The `uploaded_at` field in the `DataDocument` model has been supplanted by the `created_at` field the object inherits from the `CommonInfo` parent model. 

There are some existing records where the `uploaded_at` timestamp was assigned before the object inheritance was implemented, so their `uploaded_at` value is earlier than their `created_at` value. It might be advisable to use a migration to update `created_at` to reflect the pre-inheritance value. 

Note that `uploaded_at` is used in `dashboard\search_indexes.py`. That index will need to be changed to use `created_at` instead.

Test: Data migration test needed to check that when uploaded_at is 'older' than created_at then created_at needs to be updated to be same date as stored in uploaded_at

debboutr:
just got through creating the migrations to replace `created_at` and removing the field `uploaded_at`. I had to allow the indexed field to accept nulls because some DataDocs are this way in our seed data
`uploaded_at      = indexes.DateTimeField(model_attr='created_at', null=True)`
it seems our model for `CommonInfo` has both of the DateTime fields accepting nulls, maybe it should just be the `updated_at`?

Secondly, there are a couple of places that we used `uploaded_at` in the `views/dashboard.py` and after replacing w/ created at I am getting an error from code that starts on line 43:
```
document_stats_by_type = DataDocument.objects.extra(select=select_upload_date) \
        .values('upload_date') \
        .annotate(source_type = F('document_type__title'), document_count = Count('id')) \
        .order_by('upload_date') 
```
**here is the error:**
`OperationalError: (1052, "Column 'created_at' in field list is ambiguous")`

I think it's stemming from the **F()** statement that we make, the DocumentType model has the same field. Is there a way to specify the model that we want ? I haven't spent a lot of time on it yet, but thought someone else may have a simple solution w/o me having to muck around a bunch, you can pull branch `311` if you want to see the error

PR #405, opened 2018-09-08
311 remove uploaded at field

debboutr:
#311 
this branch now has the `uploaded_at` field removed from seed data, so load from `dev` or elsewhere to run the migration that updates `created_at` if `updated_at` is older and then the migration that removes the field from the DataDocument table.

I read in the documentation that the [`.extra()`](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#extra) method we are using on querysets in `dashboard.py` is going to be deprecated so at some point we should probably figure out a better way to handle these queries.
The `uploaded_at` field was used in `search_indexes.py` as well so make sure you can `rebuild_index`


zachmunzenrider:
As per Matt's response to my question, I changed the search index field name, and also merged Dave A's changes in. I'm not sure if I can approve this, but if I can, I do!

PR #409, opened 2018-09-11
355 product page

zachmunzenrider:
None of our test products have images, and it doesn't look like the functionality to upload images is in place (the image columns just represent urls), so I added a sample image to our static directory. Once you've verified that no image displays on the product_detail page, you can run this query on your dev database, then refresh your browser to see the image:

`update factotum.dashboard_product set medium_image = '/static/images/dave_or_grant.png'`


mwfrost:
I don't think I understand the grid layout in use. What is the 6-column center div for, and why are the descriptive fields crammed into that 3-column div on the right?


sc_2018911_155340_81

PR #425, opened 2018-09-18
adjust dict for extracted text, add pdf icon to detail table

debboutr:


dashboard/views/data_group.py
```diff
@@ -163,15 +163,14 @@ def data_group_detail(request, pk,
             good_records = []
             for i, row in enumerate(csv.DictReader(info)):
                 # first 6 columns comprise extracted_text data
+                doc = docs.get(pk=row['data_document_id'])
                 extracted_text_data = OrderedDict(islice(row.items(),6))
                 extracted_text_data.pop('data_document_filename') # not needed in dict
+                doc.raw_category = extracted_text_data.pop('raw_category')
```
Daveahern: This is good, doing the pop here before using the `extracted_text_data` dict will remove that value from the dict.  This is what was causing the error later.

dashboard/views/data_group.py
```diff
@@ -163,15 +163,14 @@ def data_group_detail(request, pk,
             good_records = []
             for i, row in enumerate(csv.DictReader(info)):
                 # first 6 columns comprise extracted_text data
+                doc = docs.get(pk=row['data_document_id'])
```
Daveahern: Slightly more efficient assignment here, removing what had previously been in lines 170 and 171.

PR #429, opened 2018-09-25
393 reformat data document page

zachmunzenrider:
I updated seed data for group types and document types as part of this ticket. Be sure to load those as part of testing!

Two good test documents are 53 (Habits and Practices) and 7 (Composition):
http://127.0.0.1:8000/datadocument/53/
http://127.0.0.1:8000/datadocument/7/


zachmunzenrider:
Dave - Thanks for testing! The grouptype=4 was a merging error on my part, and I changed it to match dev.

I also made a change to the script url display, but I think you uncovered an edge case scenario: the extraction_script for datadocument 53 (one of the documents you were testing with) is "Manual (Dummy)" and has a blank URL. All the other extraction_script urls point to actual github links. I updated the href on the page, but this is primarily a byproduct of our test data.

Issue #416, opened 2018-09-12
upload extracted text for 'chemical presence list' data group type (13 pts)


kdionisio:
I want to be able to upload extracted text for a new data group type, 'Chemical presence list'

For data group type 'Chemical presence list', the upload extracted text template should include the following fields. Uploaded data will be pushed into the extractedtext (existing) and extractedlistpresence (new) models.
- Existing fields (these exist in our data models)
  - data_document_id (extractedtext and datadocument model)
  - data_document_filename (datadocument model)
  - doc_date (extractedtext model)
  - raw_category (datadocument model)
- New fields (to go into a new extractedlistpresence model)
  - raw_cas (text, max 100 characters)
  - raw_chem_name (text, max 500 characters)
  - cat_code (text, max 100 characters)
  - description_cpcat (text, max 200 characters)
  - cpcat_code (text, max 20 characters)
  - cpcat_sourcetype (text, 50 characters)

Definition of Done:
- ability to upload extracted text for data group type 'Chemical presence list'

Test:

kdionisio:
Additional clarification notes per discussion with developer. The following fields:

cat_code (text, max 100 characters)
description_cpcat (text, max 200 characters)
cpcat_code (text, max 20 characters)
cpcat_sourcetype (text, 50 characters)

are unique to the document and not to the specific chemical. 

Thus to maximize efficiency and avoid duplication, developer can choose how best to organize/store these fields, regardless of the model that the above ticket indicates the fields should be stored in.

PR #438, opened 2018-09-28
416 chemical presence list upload

debboutr:
#416 
you can use `/datagroup/49/` for a DG to upload a template and try to submit.


debboutr:
@zachmunzenrider, we had determined on an earlier phone call that the 4 CPCat fields:

- cat_code (text, max 100 characters)
- description_cpcat (text, max 200 characters)
- cpcat_code (text, max 20 characters)
- cpcat_sourcetype (text, 50 characters)

were unique only to the document and not to the specific chemical, so Kathie had said that I could handle it however I saw fit. If you remember, we had talked about inheritance on the phone and how best to handle the associations in these tables **ExtractedText, ExtractedChemical, ExtractedFunctionalUse**. This ticket adds another model to manage with all of the data linked to a DataDocument, so for now I have just added the functionality and the tables to get what we want, i.e. uploading the data that the ticket specifies. We may at some point in the future look for a way to reorganize the data in the DB so that there aren't useless columns or duplicates across models like `report_funcuse`. I inherited The ExtractedText model when creating the ExtractedCPCat model so that we could all get familiar with that type of an association to use it if we decide to refactor these tables. Hope this helps

Issue #367, opened 2018-08-07
Data Group Edit Redirect back to Data Group (2 pts)


dmlyons2:
While on the data group view page (factotum.epa.gov/datagroup/19) and I press edit I am taken to the form to edit the data group. If I press submit or cancel I am taken to the list of data groups. I would like to be taken back to the page that I was working on, i.e. data group view page (factotum.epa.gov/datagroup/19)

Def of Done:

data group edit returns to the previous page (either data group list or data group view)

Test: data group edit returns to the previous page

PR #454, opened 2018-10-08
#367 redirect from datagroup edit page

debboutr:
we had originally had used the `request.META` dict to issue the `HTTP_REFERER` key to the href in the Cancel button, which will always work whenever the user clicks the edit button on either the datagroup_list page or the datasource_detail page or the datagroup_detail page, the addition of qualifying it through the `form.referer` attribute really just covers the case where the user would have dropped the full URL into a new tab or window, or if they were to open the link in a new tab or window. Either of these removes "HTTP_REFERER" value in the dict. So, in the event that they click Cancel in a case like this, they will be directed to the datagroup_list view.

another thing to be aware of... we have defined in our `urls.py` file the trailing `/` to most urls, but not in the case of the `data_source_detail` url, this is handled in the tests, but we might consider having consistency with each url, OR we could setup the same namespace and view function to either

```
path('datasource/<int:pk>', views.data_source_detail,
                                          name='data_source_detail'),
path('datasource/<int:pk>/', views.data_source_detail,
                                          name='data_source_detail'), 
```


debboutr:
this was approved, but I had forgotten to add the renamed test file to the stage so I've done that with this last commit.

Issue #448, opened 2018-10-04
error when registering records (2 pts)


kdionisio:
Encountered error when attempting to upload a 'register records.csv' file when creating a new data group of data group type Composition.

Error:
DataError at /datagroup/new/
(1406, "Data too long for column 'filename' at row 1")

I am unclear as to why the filename in row 1 is too long as it is only 18 characters (1_DREFT_powder.pdf) and the 'filename' field in the datadocument model can take up to 100 characters. I also registered records with longer filenames yesterday without encountering any issues.

Also note that though I saw an error, the data group was created within the data source (I thought that when an error was thrown when registering records the data group would not be created??)

[P&G_SDS_register_records.zip](https://github.com/HumanExposure/factotum/files/2447435/P.G_SDS_register_records.zip)

debboutr:
this error stems from line:
`1354_Head%20_and_%20Shoulders`

there may be a better way to validate each line of the csv w/ a `clean` method on the DataDocument object before the `.save`? 

Here are all of the records that throw an error for having a `filename` that is too long in the attached file above

![errors](https://user-images.githubusercontent.com/7052993/46517676-2e3e9a80-c825-11e8-973e-608488d7d29d.png)

the only check we do on filename currently is check to see that it is not empty `''` which would also throw a similar error when the save method is invoked.

mwfrost:
I'm in favor of using real ORM logic to validate the records rather than trying to validate csv cells straight from the source.

kdionisio:
I'm also wondering if it is easy or hard to increase the allowed length of the file name? If easy I would also request to do so (or can ticket as an additional ticket if preferred). I am dealing with lots of filenames provided to me by others and this isn't the first time I've had too long filenames I had to manipulate.

PR #458, opened 2018-10-10
448 register records bug, add length to filename in DataDocument

debboutr:
#448 

added more characters to the filename field and added test to handle a more graceful fail on validating.


dashboard/tests/functional/test_datagroup_create_upload.py
```diff
@@ -29,9 +29,10 @@ def tearDown(self):
         dg.delete()
 
     def test_datagroup_create(self):
+        long_fn = 'a filename that is too long ' * 10
```
Daveahern: I've learned something new - I don't think I ever realized that the * operator could be used on a string to essentially multiply it.  Fun stuff!

Issue #413, opened 2018-09-12
display extracted text + data for 'chemical presence list' data group type (5 pts)


kdionisio:
**Dependent on completion of #393**

On the data document view page (http://factotum.epa.gov/datadocument/1224), extracted text displayed will depend on the data group type. Display of extracted text data from certain data group types was detailed in #393 .

For data group type 'Chemical presence list', please display:

- From extractedtext model, display following fields (all fields should have ability to be edited by the user, with the exception of extraction script)
  -  prod_name
  - doc_date

- From chemicalpresencelist model, display following fields (all fields should have ability to be edited by the user)
  - raw_cas
  - raw_chem_name

mwfrost:
In the course of doing this ticket's work, I'm adding a `code` field to the `GroupType` object. This will be a short code used to perform conditional logic on data groups - an alternative to comparing long strings. 

Unfortunately, django doesn't allow `unique=True` on a character field, so we cannot enforce the uniqueness of the codes.

The `GroupType` object was previously added in migration 0049. I have moved it to 0066.

debboutr:
django does allow unique=True on CharField, check how you have `null` and `blank` defined

[https://docs.djangoproject.com/en/2.1/ref/models/fields/#unique](https://docs.djangoproject.com/en/2.1/ref/models/fields/#unique)

also, you should probably hold off on this ticket until I get done w/ the one I'm currently working on which is where the `ExtractedPresenceList` model is getting migrated in

mwfrost:
@kdionisio , should the Extracted Text section of the Data Document detail page look different when the data group is a Chemical Presence List? The _parent_ objects have a different model type, with type-specific fields like "Cat Code," "CPCat Code," and "Description CPCat." Should those appear in the box under "Extracted Text"? And should they be editable?

debboutr:
tests to write: 

- `/qa/extractedtext/pk` view needs to always have `chem-control` in it's fields for `toggleChemicalEdit` fn
- unit test the `get_extracted_models` and `create_detail_formset` functions

PR #459, opened 2018-10-10
413 Dynamic, conditional creation of inline formsets for extracted data

mwfrost:
Ticket #413 required the creation of a new inline formset for displaying all the `ExtractedChemicalPresence` records that belong to an `ExtractedCPCat` object. The `ExtractedCPCat` object inherits from `ExtractedText` but needed to be treated differently by the view and template. 

Rather than simply adding another `if:` to the branching conditional logic that creates formsets for `ExtractedChemical`, `ExtractedFunctionalUse`, and `ExtractedHabitsAndPractices`, I built the method `create_detail_formset()` inside `forms.py`. It takes an optional POST request and a required instance of the detail records' parent object: currently either an `ExtractedText` or an `ExtractedCPCat` object. Then it passes the parent object and its fields to the `inlineformset_factory()`, which collects and displays the appropriate type of child record.

The `data_document_detail.html` template currently uses this method only for the `ExtractedCPCat` / `ExtractedChemicalPresence` relationship, but other new or existing models can hopefully be plugged into the same logic in future tickets.


mwfrost:
Note: I added the method `fetch_extracted_records()` to the `ExtractedText` object. This method gathers all the children of each `ExtractedText` object, regardless of their class.

```
    def fetch_extracted_records(self):
        '''Collect the related objects in all the Extracted... models
        '''
        # Start with the known children of the base Model: ExtractedText
        full_chain = chain(self.practices.all(), 
                    self.chemicals.all(), 
                    self.uses.all()
                    )
        # Try to get all the child objects of derived (inherited) models

        # ExtractedCPCat has related ExtractedListPresence objects connected
        # by the .presence relation
        if hasattr(self, 'extractedcpcat'):
            presence_chain = self.extractedcpcat.presence.all()
            full_chain = chain(full_chain, presence_chain)

        return full_chain
```

dashboard/forms.py
```diff
@@ -30,12 +34,94 @@ class Meta:
...
+        """         print('Creating DetailFormset for group_type %s ' % dg_code)
+        print('    parent_model: %s ' % ex_model)
+        print('    detail_model: %s ' % detail_model)
+        print('    fields: %s ' % detail_fields)
+        print('    detail record count: %s' % len(list(ex.fetch_extracted_records())) ) """
+        DetailFormset = forms.inlineformset_factory(parent_model=ex_model,
+                                                    model=detail_model,
+                                                    fields=detail_fields,
+                                                    extra=1)
+        if (req):
```
debboutr: I'm getting a fail due to this `if/else` statement, if we make a GET request to get the `extracted_detail_form` object it fails. We need to resolve for `req.method == 'GET'`

dashboard/forms.py
```diff
@@ -30,12 +34,94 @@ class Meta:
...
+            if req.method == 'POST' :
+                extracted_detail_form = DetailFormset(req.POST, instance=ex, prefix='detail')
+            else:
+                extracted_detail_form = DetailFormset(instance=ex, prefix='detail')
+        else:
```
debboutr: this only applies to the use in `tests`, might be best to handle this logic (109-116) in the view.

dashboard/forms.py
```diff
@@ -30,12 +34,94 @@ class Meta:
...
+    # The model used for the formset depends on whether the 
+    # extracted text object matches a data document 
+    # The req argument takes an http request with POST
+    # data for editing and saving objects
+    ex = parent_exobject
+    ex_model = ex.__class__
+    dd = DataDocument.objects.get(pk=ex.pk)
+    dg_code = dd.data_group.group_type.code
+
+    if (dg_code == 'FU'):               # Functional use
```
debboutr: this is similar to the logic used in the `get_extracted_models` in `utils.py`, we should be able to get these to work together. I am going to branch from here and factor out some of the repeated logic. We will also need to adjust the places where we were qualifying by the title of the `GroupType`. It might make sense to drop the `title` field and add `choices` argument to `code` or vice versa. I don't think that we will ever find an instance where the same `code` would have different `titles`, but this would be something to ask about. I think adding the following to the `DataGroup` model could serve us well in many places in the code, both in views and in templates:
``` @property
    def type(self):
        return str(self.group_type.code)

    @property
    def is_composition(self):
        return self.type == 'CO'

    @property
    def is_habits_and_practices(self):
        return self.type == 'HP'

    @property
    def is_functional_use(self):
        return self.type == 'FU'```

File: dashboard/tests/functional/test_datadocument_detail.py [hunk 4]
```diff
@@ -33,4 +35,67 @@ def test_documenttype_update(self):
...
+                #print('    %s: %s' % (ex_child.__class__.__name__ , ex_child ))
+                self.assertEqual(et.pk , child_model.objects.get(pk=ex_child.pk).extracted_text.pk,
+                    'The ExtractedChemical object with the returned child pk should have the correct extracted_text parent')
+        
+    def test_every_extractedtext(self):
+        ''''Loop through all the ExtractedText objects and confirm that the new
+        create_detail_formset method returns form values that match what the old  
+        method delivered
+        '''
+        for et in ExtractedText.objects.all():
```
debboutr: this is really great that you have worked through this, but by writing this test you are pinning all of the `old_fs` definitions to remain in our code base, which is a little strange to me?

dashboard/models/extracted_text.py
```diff
@@ -36,6 +37,30 @@ def next_extracted_text_in_qa_group(self):
             nextid = 0
         return nextid
 
+    def fetch_extracted_records(self):
```
debboutr: This method I guess is just used for tests?? 2 of the 3 queries that you're chaining together are always going to be empty OR 3/4 if it's a `ListPresence`. Where is the logic that would get this test to fail? _I'm pretty certain it's not possible, but I would love to be disproven._ You're finding every child of a ExText instance and then going back and asking is this child related to this ExText, well it would have to be or it never would have been selected to begin with. And even if more than 1 query to the DB did return a queryset, it would still check out.  Do you predict this method will be useful in the code in the future?

debboutr: note you could also use the standard library to give you an iterable to loop through:
`list(self.practices.all()) + list(self.chemicals.all()) + list(self.uses.all())`

dashboard/forms.py
```diff
@@ -30,12 +34,94 @@ class Meta:
...
+        print('    detail_model: %s ' % detail_model)
+        print('    fields: %s ' % detail_fields)
+        print('    detail record count: %s' % len(list(ex.fetch_extracted_records())) ) """
+        DetailFormset = forms.inlineformset_factory(parent_model=ex_model,
+                                                    model=detail_model,
+                                                    fields=detail_fields,
+                                                    extra=1)
+        if (req):
+            if req.method == 'POST' :
+                extracted_detail_form = DetailFormset(req.POST, instance=ex, prefix='detail')
```
debboutr: we have had different prefixes in the given detail formset, but we never use them, I'd think for now that we could keep them all the same.

templates/data_document/data_document_detail.html
```diff
@@ -108,6 +117,55 @@ <h4>Extracted Text</h4>
     </div>
     </div>
 
+    <!-- This code is only for rendering the Chemical Presence data group, 
+    but the detail_formset can eventually return inline formsets for 
+    every type of data group -->
+
+    {%  if doc.data_group.group_type.code == 'CP' and detail_formset.forms %}
```
debboutr: I'm thinking out loud here, but why have the `and` there will always be a detail_formset returned even if empty, right? I like the idea of using the `code` to distinguish which formset comes through in the view. I'm seeing where you were going here, but it has yet to be completed. Couple of Q's for this template: 

1. Is it True that a data_document can only be linked to 1 Product? _it goes away in the `link_product_list` view if a Product is created for it, but it isn't constrained in the DB. 
2. we have 4 combinations of parent/child forms: (ExtractedText,ExtractedChemical),(ExtractedText,ExtractedFunctionalUse),(ExtractedText,ExtractedHabitsAndPractices),(ExtractedCPCat,ExtractedChemicalPresence). If we have one, we don't want to show cards for any of the others, right? I see that the DocumentType form will always be included.

dashboard/models/group_type.py
```diff
@@ -5,6 +5,7 @@
 class GroupType(CommonInfo):
     title = models.CharField(max_length=50)
     description = models.TextField(null=True, blank=True)
+    code = models.TextField(null=True, blank=True, max_length=10)
```
debboutr: why do you choose to make the length 10 if the code we are using are all length ==2?

templates/data_document/data_document_detail.html
```diff
@@ -29,6 +29,10 @@ <h1 class="col-12">
         <dd class="col-sm-9">
             {{ doc.data_group.group_type }}
         </dd>
+        <dt class="col-sm-3">Data Group Type Code</dt>
+        <dd class="col-sm-9">
+            {{ doc.data_group.group_type.code }}
+        </dd>
         {% if document_type_form %}
```
debboutr: this came in at an earlier commit, but when would there **not** be a `document_type_form`??

dashboard/forms.py
```diff
@@ -30,12 +34,94 @@ class Meta:
...
+
+    if (dg_code == 'FU'):               # Functional use
+        detail_model = ExtractedFunctionalUse
+        detail_fields = ['extracted_text','raw_cas',
+                        'raw_chem_name', 
+                        'report_funcuse'
+                        ]
+    elif (dg_code in ['CO','UN']):              # Composition
+        detail_model = ExtractedChemical
+        detail_fields = ['extracted_text','raw_cas',
```
debboutr: forgot `weight_fraction_type`

dashboard/forms.py
debboutr: I see now why you've chosen to do this. It's going to be tricky managing three forms in POST requests. I'm not sure the best way to go about it, but as I'm messing around here it seems like the POST data that's loaded has to come from that form,_or group of forms_, specifically, so it seems we're going to have to qualify on the name of the submit button. Or I've read we could create different views to handle each and then redirect back to `data_document_detail` after processing

dashboard/views/data_document.py
```diff
@@ -33,13 +28,30 @@ def data_document_detail(request, pk,
...
+
+    # The unified formset creation method should replace the others
+    detail_formset = create_detail_formset(request, extracted_text)
+
     chemical_formset = ChemicalFormSet(instance=extracted_text, prefix='chemicals')
     habits_and_practices_formset = HnPFormSet(instance=extracted_text, prefix='habits_and_practices')
     document_type_form = DocumentTypeForm(instance=doc)
+    
     if request.method == 'POST' and 'save_extracted_text' in request.POST:
         extracted_text_form = ExtractedTextForm(request.POST, instance=extracted_text)
```
debboutr: this comes out funky when it is supposed to be `ExtractedCPCatForm`

dashboard/models/group_type.py
mwfrost: Because we might be told to use longer codes at some point.

templates/data_document/data_document_detail.html
mwfrost: We could remove that `if` logic, sure. But I like knowing the page renders even if (due to some future change) the form-building code can’t render the document type form

mwfrost:
@debboutr Thanks for the detailed review! I’ll go back through each comment and evaluate which ones need to be addressed first. 

I will try to call into the standup today, depending on whether I’ve got phone service. Either way, these points can be handled between us over Slack with the occasional appeal to the team when I get to the hazier questions.

dashboard/forms.py
mwfrost: I like adding these properties

templates/data_document/data_document_detail.html
debboutr: Couldn't we also make that field non-nullable. That could also be used in the cleaning of the register_records.csv logic once we migrate to cleaning in the models. If we have a `document_type` of "unidentified" we probably shouldn't need nulls, but I think it would be worth asking @kdionisio or @dmlyons2  about.

dashboard/models/group_type.py
debboutr: I suppose that's fine, but doesn't it make more sense here to use the `choices` argument like we do in the `DataSource` model? There is no case where a given `code` would ever have a different `title` so why keep both in the DB?

dashboard/forms.py
debboutr: note also, that with just a little tweaking you could get back the 2 models needed in the creation of the formset by type with the `get_extracted_models` function in `utils.py`, we could add `detail_fields` to be returned in those funcs as well and just not use them in the DG detail view. you will need to remove the `.objects.get_or_create` method and just adjust in the view, that would remove everything above `if detail_model != None:`

dashboard/models/extracted_text.py
mwfrost: I do predict this model will be useful, but the alternative approach in `utils.py` also makes sense. I'll compare the approaches.

The `list(x) + list(y)` approach would return an iterable list, not an iterable of QuerySet objects. I decided it's better for this function to return QuerySet objects.

dashboard/forms.py
mwfrost: Good, this was one of my questions. Might we someday wish the detail formsets had selectors that we could use for color-coding? If it's true that there will never be mixed types, then I say we will be okay with "detail_" or something uniform as the prefix.

mwfrost:
These four URLs point to one datadocument of each GroupType:

`http://127.0.0.1:8000/datadocument/254780/`
`http://127.0.0.1:8000/datadocument/53/`
`http://127.0.0.1:8000/datadocument/167497/`
`http://127.0.0.1:8000/datadocument/5/`

debboutr:
we are unable to save new `extracted_chemical` objects if the UnitType dropdown is left at None

dashboard/tests/functional/test_datadocument_detail.py
```diff
@@ -29,8 +32,56 @@ def test_documenttype_update(self):
...
+            # Compare the model of the child formset's QuerySet to the model
+            # of the ExtractedText object's child objects
+            dd_child_model  = get_extracted_models(dd.data_group.group_type.code)[1]
+            childform_model = child_formset.__dict__.get('queryset').__dict__.get('model')
+            self.assertEqual(dd_child_model, childform_model)
+
+            
+
+
+        print('\nprint(some very important test output)')
```
dmlyons2: Nice depiction of a rabbit

PR #461, opened 2018-10-10
Update requirements.txt to use compatible versions of widgets

mwfrost:
Both of the packages that broke as a result of the Django 2.1 changes have been updated and can be automatically updated with pip.

Run `pip install --upgrade -r requirements.txt`


Daveahern:
Agreed, it's working well for me as well.

PR #472, opened 2018-10-17
445 register unique filenames

Daveahern:
Upload a register_records.csv with duplicate filenames to test.  Error will display.


dashboard/tests/functional/test_datagroup_create_upload.py
```diff
@@ -25,8 +25,9 @@ def setUp(self):
 
     def tearDown(self):
         # clean up the file system by deleting the data group object
-        dg = DataGroup.objects.get(name='Walmart MSDS Test Group')
-        dg.delete()
+        dg = DataGroup.objects.filter(name='Walmart MSDS Test Group')
+        if dg.__len__() > 0:
```
debboutr: why are you using __ methods here?? 
```suggestion
        if len(dg) > 0:
```
```suggestion for below...
        if dg.__len__() > 0:
```

Daveahern: Because as I started typing the code completion in my IDE suggested it and I just said ok?  How's that for honesty.  I'm not sure if there's much difference between one or the other, but I'm happy to make the change.

Issue #333, opened 2018-07-23
Dashboard Page: Extracted Chemical Count (3 pts)


dmlyons2:
On the Dashboard Page, in one of the two blank cards on the top of the page, add Extracted Chemical Count which is the count of all rows in the extracted chemical table.

Test:
View for dashboard returns correct extracted chemical count

PR #488, opened 2018-10-18
#333 add card for chemical count

debboutr:
#333 

this is just `ExtractedChemical` count, but we might also want count of `ExtractedFunctionalUse` or others to be incorporated?? _not sure_


mwfrost:
I merged in dev prior to review. The test is present and passing; the number displayed in the browser matches the number of records in my dev database. Approved!

PR #491, opened 2018-10-19
419 Use dynamic formset generation to populate QA page

mwfrost:
Use new formset generation method for populating the QA page for a chemical presence datadocument. 

Example QA page:
http://127.0.0.1:8000/qa/extractedtext/254780/


zachmunzenrider:
Nice use of formsets. Looks good

Issue #470, opened 2018-10-17
data document view page doesn't open when no extracted text


kdionisio:
data document detail view does not open if extracted text does not exist

mwfrost:
This can be reproduced in the seed data at `http://127.0.0.1:8000/datadocument/155324/`

mwfrost:
Test: run`data_document_detail()` on every datadocument in the seed data and confirm valid output.

PR #499, opened 2018-10-23
Ticket 470: Handle absent extracted text

mwfrost:
This branch satisfies the terms of ticket #470 : the data document detail page opens whether or not the data document has an associated ExtractedText object, and the test checkes every DataDocument object in the seed data.

There may be other cases where the code still expects an ExtractedText object and fails when one is absent, but I have not yet encountered them. The data group detail page, for example, loads fine. The QA pages are centered around the ExtractedText model rather than the DataDocument model.

### DIFF
--- a/dashboard/views/data_document.py
+++ b/dashboard/views/data_document.py
@@ -14,30 +14,44 @@ from dashboard.models import *
 def data_document_detail(request, pk,
                          template_name='data_document/data_document_detail.html'):
     doc = get_object_or_404(DataDocument, pk=pk, )
-    # TODO: this needs to account for the absence of an ExtractedText object
-    # https://github.com/HumanExposure/factotum/issues/470
-    extracted_text = ExtractedText.objects.get(data_document=doc)
     ParentForm, ChildForm = create_detail_formset(doc.data_group.type, EXTRA)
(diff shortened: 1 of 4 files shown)


zachmunzenrider:
Matt, does our seed data currently have a good datadocument for testing this?

mwfrost:
> Matt, does our seed data currently have a good datadocument for testing this?

The seed data included a mix of data documents with and without related ExtractedText objects, even before this issue. The ticket lists an example for testing in the browser.

dashboard/views/data_document.py
```diff
@@ -14,30 +14,44 @@
...
+            'detail_formset': child_formset}
+            )
+        if request.method== 'POST':
+            child_formset = ChildForm(request.POST, instance=extracted_text)
+            if child_formset.is_valid() and child_formset.has_changed():
+                child_formset.save()
+                child_formset = ChildForm(instance=extracted_text) # load extra form
+                colors = ['#d6d6a6','#dfcaa9','#d8e5bf'] * 47
+                color = (hex for hex in colors)
+                for form in child_formset.forms:
```
debboutr: won't this only color the cards on a POST? maybe dedent?

PR #508, opened 2018-10-29
476 bulk assign products to pucs

zachmunzenrider:


dashboard/views/product_curation.py
```diff
@@ -119,6 +118,32 @@ def detach_puc_from_product(request, pk):
     pp.delete()
     return redirect('product_detail', pk=p.pk)
 
+@login_required()
+def bulk_assign_puc_to_product(request, template_name=('product_curation/'
+                                                      'bulk_product_puc.html')):
+    q = safestring.mark_safe(request.GET.get('q', ''))
+    if q > '':
```
debboutr: an empty string satisfies this condition
```
In [58]: '   '>''                                                               
Out[58]: True
```

mwfrost: I just added `.lstrip()` to line 124 to treat `' '` or `'  '` etc as if they were `''`

dashboard/views/product_curation.py
```diff
@@ -119,6 +118,32 @@ def detach_puc_from_product(request, pk):
...
+                filter(brand_name__icontains=q))[:50]
+    else:
+        p = {}
+    form = ProductPUCForm(request.POST or None)
+    if form.is_valid():
+        puc = PUC.objects.get(id=form['puc'].value())
+        product_ids = safestring.mark_safe(request.POST.get('id_pks', '')).split(",")
+        for id in product_ids:
+            product = Product.objects.get(id=id)
+            producttopuc = ProductToPUC.objects.filter(product=product, classification_method='MA')
```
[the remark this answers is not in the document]
mwfrost: I think I agree with Rick. A product with an assigned PUC could only appear in `id_pks` if the PUC had been assigned by some other user after the current user ran the search query. If we want to account for that scenario, we can think about it in more detail. Otherwise, I don't think this view needs to wipe and replace any ProductToPUC objects.

Issue #460, opened 2018-10-10
Replace all numeric primary keys in uploaded file templates with character codes (5 pts)


mwfrost:
Some of the file upload templates include a column that requires the user to fill in the numeric primary key of a lookup record. One example is the register records template, which uses a `document_type` column to assign that attribute to new data documents. The primary key of the lookup value, though, is automatically assigned by the DBMS, and might not match the intended value (this is mostly a problem in test suites, since the fixtures usually specify a primary key for lookup records in yaml files).

These numbers should be replaced with codes like the ones used for the data group type. 

Also add a code look-up reference on the application for easy reference.

PR #536, opened 2018-11-01
add document codes

debboutr:
#460 

the register_records.csv now uses 2 digit ## document type codes for registration.

Take a look at the diff on migration 0049, I've had to make it access the DocumentType model a little differently than it was in order for this change to migrate, and I also make it reversible.

as for the cards that are now on the datagroup_form.html, there may be a better way of pinning them down like I have to stay below the taskbar.


debboutr:
I added an optional solution to how the code tables could be laid out in the nav bar on branch `460_nav_bar`. switch over to this branch and check it out too.

mwfrost:
Closing this request in favor of one that uses the `460_nav_bar` branch

Issue #521, opened 2018-10-31
display UPC in search results table (3 pts)


kdionisio:
Please add a column for 'UPC' in the search results table (UPC column would only be populated for products, not for data documents).

UPC allows user to uniquely identify an entry in the search results (e.g. http://factotum.epa.gov/find/?q=laundry).

Definition of Done:
UPC column appears in search results table.

Test:

PR #540, opened 2018-11-02
521 alternate format: new search results layout and css

mwfrost:
This is an alternate solution to ticket #521 that displays the UPC code in the search results as requested, but abandons the table format in favor of a list. The metadata is displayed under each object title, and filter links have been added to the metadata results (like Product/Data Document or Brand Name) where possible.


mwfrost:
You might notice that the facet-filtering links work by appending the unescaped string to the current URL. So if you are at `http://127.0.0.1:8000/find/?q=avcat` and hover on a "Data Document" link, you'll see that it links to `http://127.0.0.1:8000/find/?q=avcat&facet_model_name=Data Document`. Once you click the link, the string in "Data Document" is properly escaped and you're taken to `.../find/?q=avcat&facet_model_name=Data%20Document`. That could probably stand to be tightened up.

Also, once you get to that filtered page, there's no conditional logic applied to the link-creation tags. So when you hover over the same link this time, it points to `http://127.0.0.1:8000/find/?q=avcat&facet_model_name=Data%20Document&facet_model_name=Data%20Document`. The filtering strings just keep stacking up. It works, but it's ugly. 

If this proof-of-concept looks promising, we can clean those two things up and test them better.

zachmunzenrider:
Ah--I realize now that the upcs are "stub_xx" not "stub xx". I didn't see the underscore in my MySQL Workbench. :-)  So "stub 11" bring back results like "Eleanors VF-11 Plant Food | stub_1703", which is reasonable, though I'd think it would bring back "stub_11" too. Still, not a show-stopper! I don't think this would be a real-world issue.

I'd still like to see the other little changes, and I definitely prefer the div layout to the tabular layout!

Daveahern:
@mwfrost if we merge this PR in, does it nullify the other PR for #521 ?

mwfrost:
> @mwfrost if we merge this PR in, does it nullify the other PR for #521 ?

Yes - this was branched off a feature-complete version of the other 521 branch.

templates/search/facet_search.html
```diff
@@ -118,49 +118,66 @@ <h3>Data Document Filters</h3>
...
-                  <th>Type</th>
-                  <th>Record</th>
-                  <th>PUC</th>
-                  <th>Brand Name</th>
-                </thead>
-                <tbody>
+            <div class="results-wrapper">
+
+                {% if  page_obj.object_list %}
+
```
Daveahern: @mwfrost or @zachmunzenrider correct me if I'm wrong, but in this branch the <table> code has been removed.  But the test `test_faceted_search_renders_table` is still looking for a <table> tag and thus the test fails for me.  Am I missing somethign?

zachmunzenrider: You know, I'm not 100% sure I ran the tests on this alternate version. But you're right, this definitely shouldn't work with the non-table layout!

Daveahern: Ok why don't I change that test to AssertNotContains `table` since we're explicitly not using table for result presentation.

zachmunzenrider: Sounds good! Do you want to do that or shall I?

Daveahern: I'm in here.  I also need to get `dev` merged in since there's a conflict to resolve.  I'll give it a go.

PR #543, opened 2018-11-05
#460: switch from numeric document type keys to alpha codes

mwfrost:
I've made this PR as the reviewer. I prefer this branch to the primary branch for ticket #460, and I've merged the primary branch into it.


dashboard/tests/unit/test_models.py
```diff
@@ -16,7 +17,9 @@ def create_data_documents(data_group, source_type, pdfs):
                 line['title'] = line['filename'].split('.')[0]
             dd = DataDocument.objects.create(filename=line['filename'],
                                             title=line['title'],
-                                            document_type=DocumentType.objects.get(pk=line['document_type']),
+                                            document_type=DocumentType.objects.get(
```
debboutr: this test no longer reads the `document_type` from the CSV file, it's just forcing the type to be 'MS' w/in the give datagroup's group type, also, why do we need the `Q` object here?? the lookups would get "AND"ed together anyway

`DocumentType.objects.get(code=line['document_type'], group_type=data_group.group_type)`

mwfrost: Try removing the `Q()` and see what happens.

PR #547, opened 2018-11-06
#490 Fix by replacing `doc` in template with `script`; add a test

mwfrost:


dashboard/tests/functional/test_extracted_qa.py
```diff
@@ -23,6 +23,12 @@ def test_qa_group_creation(self):
         ext = ExtractedText.objects.get(qa_group=qa_group)
         self.assertIsNotNone(ext.qa_group)
         response = self.client.get(f'/qa/extractedtext/{ext.pk}/')
+        # The extracted text object's QA page should include an "Exit" link
+        # that goes to the extraction script's QA page
+        self.assertInHTML(
+            '<a class="btn btn-secondary btn-lg btn-block" href="/qa/extractionscript/%s/" role="button">Exit</a>' % self.objects.extext.extraction_script.pk,
```
debboutr: I like this form of test, doesn't seem like we've used it yet but it seems real handy.

PR #548, opened 2018-11-06
486 add a simple test for uploading a txt file

mwfrost:


debboutr:
didn't we just discover yesterday, the futility of the test that is like this one that uses the `create_data_documents` function? DataDocuments are made in the `data_group_create` view. In my opinion this is where these tests should be focusing, at this view and then maybe a test of the `data_group_detail` view that sees these files of a different extension are uploaded on a POST w/ the given directory of where the file is. These builder functions that we are using, i.e. `create_data_documents_with_txt` aren't worth much to our testing like we have been seeing with the `loader.py` file. They're just us using python commands to get the DB where we want it for tests, this can be built into the yaml fixtures. We've known this for a while, but it's a big undertaking. It seems the test is really concerned with the file being uploaded so should probably use the POST method of the `detail` view to see that a filename w/ .txt gets uploaded as well.

dashboard/tests/unit/test_models.py
```diff
@@ -135,6 +158,29 @@ def test_data_document_filename(self):
...
+        dg_response = self.client.get(f'/datagroup/{str(self.objects.dg.pk)}/')
+        self.assertIn(b'NUTRA', dg_response.content)
+        self.assertEqual(len(self.pdf_txt), 2)
+        # Confirm that the two data documents in the csv file are matches to
+        # the pdfs via their file names
+        self.assertEqual(self.objects.dg.matched_docs(), 2)
+        # Test a link to an uploaded pdf
+        fn = docs[0].get_abstract_filename()
+        u = "{0}/pdf/{1}".format(self.objects.dg.fs_id, fn).encode('utf-8')
+        self.assertIn(u, dg_response.content, (
```
debboutr: isn't this where we want there to be a `.txt`????

Daveahern: That's a good point.  We're testing upload of a .txt, which is being partially verified, but testing the link to the .txt would make it complete.

Daveahern:
So @debboutr and @mwfrost I updated the test to use the second [1] data document, which is the text file, for firing off and testing the link.  Tell me if you guys want to look at this before I merge it in.

Issue #551, opened 2018-11-07
remove edit functionality for extracted text data from data document page (5 pts)


kdionisio:
Due to science requirements around QA and tracking edits to data, all edits to extracted text fields will need to go through the QA workflow so that edits can be appropriately tracked.

Definition of Done:
1. Ability to edit extracted text fields on data document page (http://factotum.epa.gov/datadocument/268235/) is removed (fields remain, but are not able to be edited). Fields this applies to on the DD page include:
- In Extracted Text box
     - Product name
     - Doc date
     - Rev Num
- In Composition section
    - all fields
2. Associated 'Save composition edits' button and 'Save edits' button (in extracted text box) can be removed.

PR #566, opened 2018-11-08
#551 Data Document detail view no longer allows editing of extracted data

mwfrost:
Now the only editable field on the data document detail view is the Data Document Type dropdown.


dashboard/views/data_document.py
```diff
@@ -14,7 +14,7 @@
 def data_document_detail(request, pk,
                          template_name='data_document/data_document_detail.html'):
     doc = get_object_or_404(DataDocument, pk=pk, )
-    ParentForm, ChildForm = create_detail_formset(doc.data_group.type, EXTRA)
+    ParentForm, ChildForm = create_detail_formset(doc.data_group.type, extra=0, can_delete=False)
```
debboutr: this `can_delete` argument isn't doing anything in the code except being set to `True` or `False`, beyond that it doesn't influence the code in anyway so if it was removed it wouldn't matter. Hence the question, why add it?

dashboard/views/data_document.py
```diff
@@ -35,17 +35,19 @@ def data_document_detail(request, pk,
         extracted_text = extracted_text.pull_out_cp() #get CP if exists
         extracted_text_form = ParentForm(instance=extracted_text)
         child_formset = ChildForm(instance=extracted_text)
+        # The fields should be read-only in this view
+        for f in extracted_text_form.fields:
```
debboutr: does it make sense to keep these in forms? the user is going to see an input box that can never be used?? not a big deal, but I'd say I don't like the idea. We can add the color to the children's queryset and paint the cards that way

dashboard/views/data_document.py
mwfrost: The can_delete variable is passed to the `forms.inline_formset_factory()` method. It determines whether or not the form returns a "Delete" checkbox or not. See `forms.py`

dashboard/views/data_document.py
mwfrost: This makes sense - the template would be simpler if we removed the now-unused error and validation logic and just pulled the ExtractedText objects from the ORM.

dashboard/views/data_document.py
debboutr: ahaa! thank you for the answer, I reviewed this on github w/o pulling or looking at anything else but the diff, foolish, yes, but another good reason to remove this view from using forms.

dashboard/views/data_document.py
mwfrost: @debboutr See what you think of the latest version.

debboutr: this looks good so far, we should be able to use `get_extracted_models` in utils.py to pick up the 2 models for querying and loading up the cards

zachmunzenrider:
The code seems to do what it needs to for this ticket, so that's all good. I ran into a test failure that turned out to stem from a change in the capitalization of a label. I standardized the label capitalization, which fixed the test, and pushed that change back up.

A quick review of the system suggests we're being totally inconsistent with our capitalization of words in labels. In the interests of professionalism, we should probably come to an agreement about this and try to be consistent!

dmlyons2:
@mwfrost - When you have a chance please resolve the conflicts on this branch in the `templates/data_document/data_document_detail.html` file and push with conflicts solved

PR #567, opened 2018-11-13
refactor, clean dashboard

debboutr:
this isn't an issue, but I found some ways to clean up the dashboard view that I think make the code a little easier to read through. There was a lot of redundancy in the `index` view and there can probably still be more done, but hopefully what I've done here will be a little easier to read through. 

This doesn't ever have to be merged, I've just always looked at this page and been confused, so at least it was a way for me to work through everything that is going on here.


dashboard/views/dashboard.py
```diff
@@ -6,108 +6,82 @@
...
 def datadocument_count_by_date():
     # Datasets to populate linechart with document-upload statistics
-    # Number of datadocuments, both overall and by type, that have been uploaded as of each date
-    select_upload_date = {"upload_date": """date(dashboard_datadocument.created_at)"""}
-    document_stats = {}
-    document_stats['all'] = list(DataDocument.objects.extra(select=select_upload_date) \
-                                 .values('upload_date') \
-                                 .annotate(document_count = Count('id')) \
-                                 .order_by('upload_date'))
-    document_stats_by_type = DataDocument.objects.extra(select=select_upload_date) \
```
debboutr: lines 47-52 should have been pulled out in commit `0b1c95c`, where we removed the stats by type in the chart.

mwfrost:
I'm closing this because dev has gotten way ahead of these improvements. If it's worth refactoring the code (and it might be), we should open a ticket and use this for reference.

PR #576, opened 2018-11-15
519 enhancement stub upc products

wilmelz:
Added ability to run script as a management command.  Added YAML type output.  Added writing of PKs to output.

### DIFF
--- a/factotum/management/commands/product_dup_check.py
+++ b/factotum/management/commands/product_dup_check.py
@@ -1,0 +1,83 @@
+import django
+import os
+import yaml
+
+from django.core.management.base import BaseCommand
+
+# settings need setup before calling any Django code
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "factotum.settings")
+django.setup()
+
+from dashboard.models import Product
+
+# Import the fuzzywuzzy package/libraries.
+from fuzzywuzzy import fuzz
+from fuzzywuzzy import process
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **kwargs):
+        help = 'Score stub product titles'
+
+        # Get all the products that have stub UPCs.
+        stub_products = Product.objects.filter(upc__contains="stub")
+
+        # Create a variable that will have all products that have non-stub UPCs.
+        norm_products = Product.objects.exclude(upc__contains="stub")
+
+        # Open file that will contain the top match for each stub product
+        file_topmatch = open("top_match.txt", "w")
+
+        # Open file that will contain matching scores for the process.extractOne method
+        file_multimatches = open("multi_matches.txt", "w")
+
+        file_yaml = open("document.yaml", 'w')
+
+        # Set the number of matches ordered by scores.  By default this number is 5.
+        num_matches = 3
+
+        # Set the threshold value to use for the matching score.  This value will be used for each stub product
+        # that is compared to every UPC product. Adjust this parameter to a desired threshold. This parameter has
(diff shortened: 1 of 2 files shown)


debboutr:
this runs using the command:
* `python manage.py product_dup_check`
* need to have the package `fuzzywuzzy` installed so we should add to README
I got this warning when running, so we might add this pkg as well
```
/home/debboutr/anaconda3/envs/hem/lib/python3.7/site-packages/fuzzywuzzy/fuzz.py:11: UserWarning: Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning
  warnings.warn('Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning')
```

wilmelz:
Added new package called ruamel.yaml.  This package can be installed by entering "pip install ruamel.yaml" from the terminal window. It is used for the better handling of YAML output.  Disabled warning regarding the installation of the Levenshtein package.  Added printing of output location and names of files.  Simplified the coding in the script.

debboutr:
do we want to return the product in the event that there are no matching UPC's?  how are these files going to be used?
```
- aquasorb; pk = 1564:
  - UPC Product:
    - No matches found.
- lithiun manganese dioxide batteries; pk = 1563:
  - UPC Product:
    - No matches found.
- ? "purex liquid laundry detergent \u2013 free & clear, he\u2013free & clear, original\
    \ fresh, mountain breeze, after the rain, he \u2013 after the rain, lavender (sunshine\
    \ clean canada), lemon fresh, baby (purex baby soft); pk = 1562"
  : - UPC Product:
      - No matches found.
```

wilmelz:
Added the fuzzywuzzy and ruamel.yaml packages to the requirements.txt file.

factotum/management/commands/product_dup_check.py
```diff
@@ -0,0 +1,133 @@
...
+            norm_prod = norm_product.title.lower()
+            list_upc_names.append(norm_prod)
+            pk_product = norm_product.pk
+            list_upc_pks.append(pk_product)
+
+        # Get current working directory
+        pathname = os.getcwd()
+
+        # Open file that will contain the top match for each stub product
+        file_top_yaml = open(pathname + "\\top_match.yaml", 'w')
```
mwfrost: Hard-coding the backslashes causes problems for Linux/Mac users. There's a python package we found for building filesystem directories in a platform-independent way. I'll look up an example of where it's used.

mwfrost: Found it: https://docs.python.org/3/library/pathlib.html

The `PurePath()` method builds platform-appropriate paths out of strings. I have not used it, but this would be the appropriate place.

dashboard/tests/functional/test_product_dup_check.py
```diff
@@ -0,0 +1,59 @@
...
+from fuzzywuzzy import process
+
+
+class TestProductDupCheck(TestCase):
+
+    def setUp(self):
+        # Build list of UPC product names
+        list_upc_names = []
+        list_upc_pks = []
+        for norm_product in norm_products:
```
Daveahern: @wilmelz Take a look a this. The test is failing here because `norm_products` doesn't exist yet, it's instantiated at line 33.

wilmelz: @Daveahern: Fixed the bug in test_product_dup_check.py and pushed the code changes into GitHub.

Issue #537, opened 2018-11-01
delete extracted chemical record malfunctioning (5 pts)


kdionisio:
On the QA page for a data document's extracted text, there is a tick box to 'delete' an extracted chemical record. When the tick boxes are selected and the user hits 'save' nothing appears to happen on the screen (the chemical records are still present). However if the user exits out of the process and returns to the same QA check page, the records are then gone.

Definition of Done:
If 'delete' box is ticked on QA page (http://factotum.epa.gov/qa/extractedtext/155219/), then upon hitting save button those chemical records with 'delete' ticked should disappear from the display

mwfrost:
I reproduced this on my dev environment.

PR #580, opened 2018-11-16
#537: Display results of editing extracted records immediately after saving

mwfrost:
When a user edited or deleted or added a child record on the QA page, the changes would not be immediately visible after saving. The view has been fixed to pass the updated values back in the form objects.


debboutr:
we're real close here, but the ExtractedText/ExtractedCPCat isn't being enabled/disabled, so can be changed but the save button is still disabled. If you change `chem-control` to `detail-control` in `/templates/core/bs4 _form_dl.html` I think that gets us to where we want to be. OR we might also be able to override the `__init__` on `forms.ModelForm` class like you did with the `BaseExtractedDetailFormSet`. The first I just tried and works

mwfrost:
> we're real close here, but the ExtractedText/ExtractedCPCat isn't being enabled/disabled, so can be changed but the save button is still disabled. If you change `chem-control` to `detail-control` in `/templates/core/bs4 _form_dl.html` I think that gets us to where we want to be. OR we might also be able to override the `__init__` on `forms.ModelForm` class like you did with the `BaseExtractedDetailFormSet`. The first I just tried and works

I think `chem-control` should be generalized to `detail-control` in every instance. It was originally added as a class when the only detail records were ExtractedChemical.

PR #586, opened 2018-11-21
568 bubble hierachy

debboutr:
this changes the value of `num_prods` in the downloaded PUC csv from the get data page to count all children of the over-arching category. It also adds a legend in the bubble chart.


debboutr:
@mwfrost I'm glad you added that bit to it, but personally, I think it would be better if we were going to add sorting, to put it's functionality into a button like we did w/ split and combine. I'd say it's worth having Dave flop over to your branch to see while we're doing the review, but it's not really a part of this ticket, just something that you wanted to see.

mwfrost:
I don't want to leave the PR hanging.

PR #587, opened 2018-11-21
verify prod_name is the same for every data_document

debboutr:
this is the simplest way I could think of to check that each record for a given `data_document_id` has the same value for `prod_name`. The error was being thrown because there ended up being ~224 `ExtractedChemical` cards on the QA page, each card has something like 7 `<input>` fields and Django has a default setting of a max number of 1000. 

This seems to be a unique case w/ this group of datadocuments, but to resolve @kdionisio and I thought it would be best to make sure the `data_document_id` and `prod_name` fields are the same. The picture I attached to the issue might help illustrate.

I also added to an existing test that an error will be thrown in this event. A thing to note was that I removed the `08_extractedtext.yaml`. With that being loaded into the test I realized that the test wasn't created new `ExtractedText` objects, but updating existing.

### DIFF
--- a/dashboard/tests/functional/test_extracted_text_upload.py
+++ b/dashboard/tests/functional/test_extracted_text_upload.py
@@ -14,9 +14,7 @@ from dashboard.models import *
 class UploadExtractedFileTest(TestCase):
     fixtures = ['00_superuser.yaml', '01_lookups.yaml',
                 '02_datasource.yaml', '03_datagroup.yaml', '04_PUC.yaml',
-                '05_product.yaml', '06_datadocument.yaml', '07_script.yaml',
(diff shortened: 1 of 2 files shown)


zachmunzenrider:
I agree with this fix to the issue at hand and approve this PR. 

As an additional fix to the original problem Kathie reported, we might want to add a DATA_UPLOAD_MAX_NUMBER_FIELDS setting to our settings file. Django defaults it to 1000, but if we're possibly going to have situations where we post more than this number of fields, we could up it to 10,000 or something:

https://docs.djangoproject.com/en/2.1/ref/settings/#data-upload-max-number-fields

Issue #579, opened 2018-11-16
More than one QA group causing error


kdionisio:
When hitting the 'Continue QA' button (http://factotum.epa.gov/qa/) for the Walmarts MSDS (Extract) extraction script QA group, following error was seen:

MultipleObjectsReturned at /qa/extractionscript/12/
get() returned more than one QAGroup -- it returned 4!

This occurred when a student was working on QA for this QA group, then accidentally closed a window. Upon attempting to reenter QA for this group, this was the message seen.

**Note error is also happening with the Home Depot extraction script.

**Updated to note - I checked the prod DB and confirmed that for these two extraction scripts there are multiple QA groups associated with each one. There should only be one QA group associated with each extraction script. Are we able to figure out why/how there were multiple QA groups created for a single extraction script, so that we can prevent this from happening again? Also, what is the best way to resolve this while not corrupting the DB? Note that the two scripts where we have this issue do not have any QA completed so if we wipe the QA group (for those two scripts only) and start over that is OK.

mwfrost:
These SQL commands should assign all the extracted text records within each set of duplicates to a single QA Group:

`UPDATE dashboard_extractedtext SET qa_group_id = 4 WHERE qa_group_id IN (SELECT id from dashboard_qagroup WHERE extraction_script_id = 12);`

`UPDATE dashboard_extractedtext SET qa_group_id = 2 WHERE qa_group_id IN (SELECT id from dashboard_qagroup WHERE extraction_script_id = 9);`

Then the extra QA Groups can be deleted:

`DELETE FROM dashboard_qagroup  WHERE extraction_script_id = 12 AND id <> 4;`

`DELETE FROM dashboard_qagroup  WHERE extraction_script_id = 9 AND id <> 2;`

mwfrost:
How to reproduce the error in a dev environment:

Find a datadocument/extractedtext record without an associated QA group:
```
SELECT * from dashboard_extractedtext 
where extraction_script_id 
NOT IN (SELECT extraction_script_id from dashboard_qagroup)
AND qa_group_id is null
;
```
Note that both the Script and the DataDocument models have foreign key relationships to the QA Group model. 
Document 156051 is associated with ExtractionScript 6.
Go to that document's detail page:
`http://127.0.0.1:8000/datadocument/156051/`

Click the "QA" button to go to the ExtractedText QA page:
`http://127.0.0.1:8000/datadocument/qa/extractedtext/156051/`

[...]

PR #593, opened 2018-11-29
#579 Fix handling of multiple GQ Groups

mwfrost:
See comment on already-merged code:

https://github.com/HumanExposure/factotum/commit/b5897b533a75670f0bc7b9ebc68e929b637720ad#commitcomment-31490016

### DIFF
--- a/dashboard/models/script.py
+++ b/dashboard/models/script.py
@@ -75,10 +75,15 @@ class Script(CommonInfo):
         from .qa_group import QAGroup
         from .extracted_text import ExtractedText
         es = self
-        # Confirm that there's no existing QA Group for this Script
-        if QAGroup.objects.filter(extraction_script = es).count() >= 1:
-            print('QA Group already found for %s Script' % es )
+        # Handle cases where a QA group already exists for the script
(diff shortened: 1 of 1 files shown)


dashboard/models/script.py
```diff
@@ -75,10 +75,15 @@ def create_qa_group(self, force_doc_id=None):
...
+            # This is a valid state
             return QAGroup.objects.get(extraction_script = es)
+        elif QAGroup.objects.filter(extraction_script = es).count() > 1:
+            # this is a failure mode induced by the system's allowing
+            # duplicate QA Groups to be created for a single script
+            return QAGroup.objects.filter(extraction_script = es).first()
+
         
         # Create a new QA Group for the ExtractionScript es
         qa_group = QAGroup.objects.create(extraction_script=es)
```
debboutr: seems like it might be better to return at this line and then move the code below into a method on the `QAGroup` model that could be called `assign_documents` or something..? Everything below can be dealt with on the QAGroup. Also, shouldn't the `force_doc_id` be usable to add one new doc to the group? I don't see how that gets handled.

mwfrost: Can you propose a test for this scenario?

PR #595, opened 2018-12-02
504 upload clean composition data new

zachmunzenrider:
Created 2 sample csv files--clean_comp_data_valid.csv and clean_comp_data_invalid.csv. Can be tested with datagroup 6.


dashboard/views/data_group.py
```diff
@@ -175,6 +180,56 @@ def data_group_detail(request, pk,
...
+                except ValidationError as e:
+                    context['clean_comp_err'][i+1] = e.message_dict
+                good_records.append(ingredient)
+            if context['clean_comp_err']: # if errors, send back with errors
+                context['clean_comp_data_form'].collapsed = False
+                return render(request, template_name, context)
+            if not context['clean_comp_err']:  # no saving until all errors are removed
+                for ingredient in good_records:
+                    ingredient.save()
+                fs = FileSystemStorage(store)
```
mwfrost: Storing the csv in the filesystem was not part of the ticketed requirements, so this functionality can be removed.

dashboard/forms.py
```diff
@@ -200,6 +214,16 @@ def include_extract_form(dg):
     else:
         return False
 
+def include_clean_comp_data_form(dg):
+    '''Returns the CleanCompDataForm based on conditions of DataGroup
+    type = Composition and at least 1 document extracted
+    '''
+    if not dg.type in ['CO']:
+        return False
+    if dg.extracted_docs() > 0:
```
mwfrost:
```suggestion
    if dg.extracted_docs() > 0:
```
Should this also check for ExtractedChemical records that don't already have related Ingredient records? I think the form is created and presented when the clean composition data has already been uploaded

dashboard/views/data_group.py
```diff
@@ -175,6 +180,56 @@ def data_group_detail(request, pk,
...
+                    ingredient.full_clean()
+                except ValidationError as e:
+                    context['clean_comp_err'][i+1] = e.message_dict
+                good_records.append(ingredient)
+            if context['clean_comp_err']: # if errors, send back with errors
+                context['clean_comp_data_form'].collapsed = False
+                return render(request, template_name, context)
+            if not context['clean_comp_err']:  # no saving until all errors are removed
+                for ingredient in good_records:
+                    ingredient.save()
```
mwfrost: The way this is built, every upload will replace any existing records that have the matching `extracted_chemical_id`.  That seems good to me, but I wanted to make sure it's the intended behavior.

zachmunzenrider: This was definitely my intention, based on my interpretation of the ticket. We could run it by Kathie if you think it's warranted

dashboard/forms.py
zachmunzenrider: The ticket doesn't discuss this condition--I interpreted it to mean that the user could always upload a new .csv and overwrite the existing ingredient record

dashboard/views/data_group.py
mwfrost: Not necessary.

Issue #570, opened 2018-11-14
change QA notes field to be text not char


kdionisio:
In the QA notes field (visible here http://factotum.epa.gov/qa/extractedtext/155284/), user has need to enter more than 255 characters into QA notes.

Definition of Done:
Change QA notes field to be text field rather than char.

Issue #556, opened 2018-11-07
ability to save QA notes even if 'Approve' button is not used (8 pts)


kdionisio:
Currently on the QA page (http://factotum.epa.gov/qa/extractedtext/155284/), user can not save any QA notes unless they use the 'approve' button to approve the document's QA.

Definition of Done:
edits to 'QA notes' field should be saved when the 'save edits' button is used to save edits to the extracted text, or via another 'save QA notes' button; user should not be required to 'approve' in order to save QA notes (thus QA notes should be independent of flag indicating that changes have been made to extracted text)

mwfrost:
I may have just fixed this in ticket #570 

The "Save" button now saves changes to the QANote.qa_notes value.

PR #605, opened 2018-12-04
570 Change QA Notes to long text field

mwfrost:
The `QANote.qa_notes` attribute was a CharField of limited length, and ticket #570 requested changing it to a TextField to allow longer annotation. 

I noticed that clicking the "Save" button on the `/qa/extractedtext/x` page was failing to save changes made to the qa_notes box, so I changed that view's `POST` functionality.

I also noticed that the `QANote` model was related to the `ExtractedText` model via a many-to-one ForeignKey field. This would make sense if we had built the interface to record multiple notes per ExtractedText object, like comments on a blog post, but the user interface only presents and saves a single note per ExtractedText. This branch changed the ForeignKey to a OneToOneField, so users are prohibited from adding a QANote record to the database table if there is already one with the same `extracted_text_id`.


mwfrost:
When you check this ticket, can you please confirm that it also resolves the problem described in #556  ?

PR #619, opened 2018-12-11
remove null=True from PUC model fields

debboutr:
we had the ability for a null field in `prod_fam` and `prod_type` to be either NULL or an empty string `''`. I removed the ability for the field to be nullable as is mentioned in the [docs](https://docs.djangoproject.com/en/1.10/ref/models/fields/#null), where is says to avoid using `null=True` on CharFields.

as for giving the field a default value of `''`, this was just done so that we don't have to create a `RunPython` migration to convert all of the NULLs to empty strings.


debboutr:
This needs to be migrated when brought in! The fix that I'm proposing here is to enforce that any `prod_fam` OR `prod_type` will be empty strings and not NULL. 

We could go the other way if that makes more sense...this was just the way that we had it in the seed data and was the reason that we weren't able to reproduce the error.

I had made some of the fields null in the seeddata PUCs and then run the migration and saw that it updated those fields to empty strings so that should do the trick, but if not we may have to do it using SQL.

mwfrost:
I have pushed a commit that should resolve what @dmlyons2 refers to in his comment.

PR #625, opened 2018-12-14
601 products per puc

mwfrost:
This ticket adds a new "PUCs" template and view, available from the "Data Curation" pulldown menu. It shows all the PUCs and the number of products linked to each one. The Admin screen also shows this stat on the list view.


dashboard/admin.py
```diff
@@ -1,26 +1,33 @@
...
-        fields = ['gen_cat', 'prod_fam', 'prod_type', 'description','tags']
+        fields = ['gen_cat', 'prod_fam', 'prod_type', 'description','tags',]
+        readonly_fields = ('num_products',)
         widgets = {
             'tags': LabelWidget(model=PUCTag),
         }
 
 class PUCAdmin(admin.ModelAdmin):
-    list_display = ('__str__', 'tag_list')
+    list_display = ('__str__', 'tag_list','num_products')
```
Daveahern: So we're both creating a new PUC list page that has the counts on it under data curation, but we can also see the counts on the PUC admin page.  Like that.

PR #628, opened 2018-12-18
598 batch assign attributes to products

zachmunzenrider:


dashboard/forms.py
```diff
@@ -140,31 +140,46 @@ def __init__(self, *args, **kwargs):
...
                              required=True)
     class Meta:
         model = ProductToPUC
-        fields = ['puc', 'id_pks']
+        fields = ['PUC', 'id_pks']
 
-class HabitsPUCForm(BasePUCForm):
+class BulkProductTagForm(BasePUCForm):
+    required_css_class = 'required' # adds to label tag
+    tag = forms.ModelChoiceField(queryset=PUCTag.objects.none(),
```
debboutr: why use `none()` here, shouldn't we use `all()` or some `filter()`

dashboard/views/product_curation.py
```diff
@@ -123,6 +123,32 @@ def detach_puc_from_product(request, pk):
...
+    msg = ''
+    if form['PUC'].value():
+        puc = PUC.objects.get(pk = form['PUC'].value())
+        form.fields['tag'].queryset = PUCTag.objects.filter(id__in=(PUCToTag.objects.
+                           filter(content_object=puc).
+                           values_list('tag', flat=True)))
+        products = (Product.objects.
+                    filter(id__in=(ProductToPUC.objects.filter(PUC = puc).values_list('product_id', flat=True))))
+    else:
+        products = {}
```
debboutr: having this initialized with an empty dictionary leaves the page blank with no option to select a product

templates/product_curation/bulk_product_tag.html
```diff
@@ -0,0 +1,177 @@
...
+            <td>{{ product.brand_name }}</td>
+            <td>{{ product.pk }}</td>
+        </tr>
+      {% endfor %}
+      </tbody>
+    </table>
+    <br><br>
+        {{ form.id_pks }}
+        <button type="submit" name="save" id="btn-assign-puc" class="btn btn-primary">Submit</button>
+    {% elif form.puc.value > '' %}
```
debboutr: this is cool, I never knew strings had `__lt__` and `__gt__` dunder methods
```
In [6]: 'x'<'x'
Out[6]: False

In [7]: 'x'<'y'
Out[7]: True

In [8]: 'x'<'w'
Out[8]: False

In [9]: if "":
   ...:     print('yay')
   ...:
```
an empty string evals to `False` in and if statement if that is all we are trying to pull out, whether or not the string is empty or not. Potential edge case
```
In [11]: bool('')
Out[11]: False

In [12]: bool(' ')
Out[12]: True

In [13]: ' ' > ''
Out[13]: True
```
but based on how the dropdown input works with the autocomplete, I don't think it's even possible to enter an empty string, a user can only select an option in the dropdown.

dashboard/views/product_curation.py
zachmunzenrider: The empty dictionary wasn't the problem, as it turns out; the problem with with case-sensitivity issues with the PUC/puc form field. I have no idea how this got past me--I'd swear it was working fine when I pushed it up, but sure enough, it wasn't working fine!

dashboard/forms.py
zachmunzenrider: It's really intended to set it up as an empty placeholder, which we can then apply a non-empty queryset to once the PUC is selected. I don't think it's a problem

dashboard/views/product_curation.py
```diff
@@ -123,6 +123,32 @@ def detach_puc_from_product(request, pk):
...
 
+@login_required()
+def bulk_assign_tag_to_products(request, template_name=('product_curation/'
+                                                      'bulk_product_tag.html')):
+    form = BulkProductTagForm(request.POST or None)
+    msg = ''
+    if form['puc'].value():
+        puc = PUC.objects.get(pk = form['puc'].value())
+        form.fields['tag'].queryset = PUCTag.objects.filter(id__in=(PUCToTag.objects.
+                           filter(content_object=puc).
```
debboutr: might it be better to have named the `content_object` field `puc` in this model? it's a little weird when looking in MySQL at the table and seeing a column of ids that are labelled `content_object` when by definition they are all going to be PUC ids. Same thing for the `ProductToTag` model. Is there some necessity for using `content_object`?

mwfrost: > We may have already talked about this, but can we use the term Tag instead of Attributes?

No. We lobbied for this but the term "attribute" carries semantic meaning for the users.

mwfrost:
> We may have already talked about this, but can we use the term Tag instead of Attributes?

No. We lobbied for this but the term "attribute" carries semantic meaning for the users.

dashboard/views/product_curation.py
mwfrost: If the taggit library refers to the `content_object` field by name in its internal code, then we could break things by changing the field name in a table that's been adapted from `taggit_taggeditem`. I don't know if this is the case or not.

PR #653, opened 2019-02-04
627 Move SID from DSSToxSubstance into DSSToxLookup

mwfrost:
To identify a curated chemical's authoritative SID, True CAS and True Chemical Name, the data model previously used the `DSSToxSubstance` table to store a denormalized list of RID-to-SID matches. This branch moves the SID out of that table and into the `RawChem` abstract base class. The SID attribute is now a foreign key to the normalized `DSSToxLookup` table. 

The same data curation business logic can now apply to all the `Extracted[x]` models that inherit from `RawChem`:
- `ExtractedChemical`
- `ExtractedFunctionalUse`
- `ExtractedCPCat`

### django-model-utils

This branch adds a library to `requirements.txt`, so please install `django-model-utils` in your environment before testing.

https://django-model-utils.readthedocs.io/en/latest/index.html

This library offers a new model manager called InheritanceManager that provides convenient functionality for querying heterogeneous sets of inherited objects. Rather than querying and selecting `ExtractedText` objects and then testing for whether each one belongs to the `ExtractedCPCat` class, the `select_subclasses()` and `get_subclasses()` methods allow us to return all the objects in their true subclass form. This branch removes the conditional logic that had been used to replace `ExtractedText` parent classes with their `ExtractedCPCat` child classes where applicable. A future ticket will replace the comparable business logic for all the `Extracted[x]` types.


dashboard/models/extracted_chemical.py
```diff
@@ -41,11 +41,15 @@ def __str__(self):
     def detail_fields(cls):
         return ['extracted_text','raw_cas','raw_chem_name','raw_min_comp',
             'raw_max_comp', 'unit_type','weight_fraction_type','report_funcuse',
-            'ingredient_rank','raw_central_comp']
+            'ingredient_rank','raw_central_comp','rawchem_ptr']
```
debboutr: seems weird to me that you would have to add `rawchem_ptr` here but not in the other models that inherit it, did you check it across all the datagroup types? Also, if it is only going in the form as a hidden field might it make more sense to exclude it from the `ExtractedChemicalForm` on line 240 of `forms.py` or does it serve some other purpose?

mwfrost: Other forms that don't go through editing and validation don't complain about the absence of the `rawchem_ptr` field. The QA page needs it because its view performs validation.

dashboard/migrations/0084_move_sid_to_rawchem.py
```diff
@@ -0,0 +1,28 @@
...
+            model_name='dsstoxsubstance',
+            name='rawchem_ptr',
+            field=models.OneToOneField(on_delete=django.db.models.deletion.CASCADE, related_name='old_curated_chemical', to='dashboard.RawChem'),
+        ),
+        migrations.RunSQL(
+            """
+            UPDATE dashboard_rawchem rc 
+            INNER JOIN  
+            dashboard_dsstoxlookup dss on 
+            rc.raw_chem_name = dss.true_chemname
```
Daveahern: Isn't the concept here that the raw chem name can and will sometimes be different than the true chem name?  If we have curated chemicals that need to get moved to this new structure that differ here (as they could) this query won't correctly link them to the right dsstoxlookup record.

Daveahern: Also when do the all the canonical true chem records get migrated from dsstoxsubstance to dsstoxlookup?  Am I overlooking it?

mwfrost: Both good points - I have been using the fixture that holds the lookups to import the canonical DSSToxLookup records, but there should be a different data migration to do that on production.

I'm going to modify this migration so that it populates DSSToxLookup first, then writes the RawChem keys without relying on the raw_chem_name to true_chem_name match. That was okay for testing with seed data, but _bad_ for production.

Daveahern: Right.  Sounds like a good plan.

Daveahern:
@dmlyons2 I've reviewed and approved this PR.  It's a blocker to a bunch of downstream work and I also don't want to merge it in, having reviewed it, so I want to queue it up for your attention as soon as you're back in the office on Monday.

PR #662, opened 2019-02-07
550 undo chemical curation

mwfrost:
This branch also uses the new tools available in django-model-utils. The FieldTracker class monitors the raw_cas and raw_chem_name fields for changes.


kdionisio:
@mwfrost @debboutr this is a good point and you have picked up on a slight nuance due to the changes we've made ....if an extracted chemical record is deleted, then obviously the linkage to the dsstoxsubstance would be deleted. However whether or not the related dsstoxsubstance record is deleted would depend on if that dsstoxsubstance record is also linked to other extracted chemical records. If it is, you would not want to delete the dsstoxsubstance record. If it is not, you could delete the dsstoxsubstance record.

mwfrost:
@kdionisio I am reluctant to automatically delete DSSToxLookup records. Would you be okay if we modified the admin panel to allow users to filter on all DSSToxLookup records that are not matched to RawChem records? We would put that in its own ticket.

kdionisio:
@mwfrost Yes, I'm totallly fine with this, as long as it's ok that this may mean we end up with 'orphan' DSSToxLookup records (ie DSSToxLookup records that have no associations with any extracted chemical records). I agree that in this case we would just need some way to periodically 'clean up' the DSSToxLookup table to eliminate these orphan records.

zachmunzenrider:
Working through the code, I wasn't seeing how it implements the story requirement to delete the associated dsstoxsubstance record. I'm now seeing that Matt and Kathie came to an agreement in the PR comments that this is no longer a requirement of the ticket. I think that's a fine call, and I'm happy to approve the PR at this point. I do wonder, however, if we're not following best practices by not updating the ticket itself when the requirements of the ticket change. It seems like we probably should have updated the ticket itself to reflect this new understanding. Anyone have thoughts on that?

mwfrost:
The ticket as written became obsolete once we changed the DSSToxSubstance table to a DSSToxLookup table. I guess the question becomes whether we want to use the PR or the ticket to record any changes to the specifications like that. I think the ticket makes more sense.

zachmunzenrider:
Agreed, Matt. I think someone ought to be able to browse the completed tickets themselves and get an accurate representation of what the ticket actually did. The PR is too far removed!

PR #663, opened 2019-02-07
533 Replace extraction script's internal link with external URL

mwfrost:
Add test


dashboard/tests/functional/test_qa.py
```diff
@@ -19,6 +19,13 @@ def test_qa_scoreboard(self):
         self.assertEqual(scriptcount, row_count, ('The seed data contains 1 '
                                                   'Script object with the script_type'
                                                   'EX, which should appear in this table'))
+
+        script_url = response_html.xpath(
+            '//*[@id="extraction_script_table"]/tbody/tr[' + str(row_count) + ']/td[1]/a/@href')[0]
+        self.assertEqual(script_url, 'http://www.epa.gov/', 'The URL on the page should be the external link to the script.')
```
Daveahern: This confused me for a bit because the seed data in the fixtures does not contain the www.epa.gov url.  But this method is using loader.py to load a small set of data, in which this is the correct url for the one extraction script record created.

mwfrost: A test that loops through all the scripts in the seed data and tries to open all the URLs would have caught this. Might be worth adding someday.

I've made the addition of the hyperlink conditional on the Script object having a URL.

PR #669, opened 2019-02-12
651 fix failing test

wilmelz:
This branch fixes a bug that was causing a couple of fuzzywuzzy tests to fail.


dashboard/tests/functional/test_product_dup_check.py
```diff
@@ -1,45 +1,46 @@
...
+        self.list_upc_names = []
+        self.list_upc_pks = []
         for norm_product in norm_products:
             norm_prod = norm_product.title.lower()
-            list_upc_names.append(norm_prod)
+            self.list_upc_names.append(norm_prod)
             pk_product = norm_product.pk
-            list_upc_pks.append(pk_product)
-
+            self.list_upc_pks.append(pk_product)
```
debboutr: why do we find `self.list_upc_pks`?? it is never used in the tests anywhere.

Also, what does the call above accomplish?
`@override_settings(ALLOWED_HOSTS=['testserver'])` 

I have pretty much the same fix in a PR that I did, so I'd say this is fine.

mwfrost: the allowed hosts override is left over from some early experiments, I think

Issue #666, opened 2019-02-11
error when uploading extracted text


kdionisio:
When attempting to upload extracted text for a data group type 'Chemical Presence List' (CP), the following error was seen. Note that this is the first time we have attempted loading extracted text for the CP data group. We have tried a few different extracted text files and all result in the same error.

UnboundLocalError at /datagroup/90/
local variable 'record' referenced before assignment
Request Method:	POST
Request URL:	http://factotum.epa.gov/datagroup/90/

Django Version:	2.1.2
Exception Type:	UnboundLocalError
Exception Value:	local variable 'record' referenced before assignment
Exception Location:	/data/code/factotum/dashboard/views/data_group.py in data_group_detail, line 136

kdionisio:
File user is attempting to upload, which is causing the above error (uploaded here as zip file but is being uploaded to Factotum as a .csv) . User is attempting to upload the attached file as extracted text for datagroup/90 on prod.

[Arizona_extract_template_2007-2010.zip](https://github.com/HumanExposure/factotum/files/2852098/Arizona_extract_template_2007-2010.zip)

mwfrost:
I think one of the csv records is causing a validation error, and we are not handing that error properly.
This is in `views.data_group.py`
```
                try:
                    ext.full_clean()
                    ext.save()
                    record = ext_child(**row)
                    record.full_clean()
                except ValidationError as e:
                    context['ext_err'][i+1] = e.message_dict
                good_records.append((d,ext,record))
```
The `record` variable is assigned inside the `try:` block. If there's an exception, it adds the error to `ext_err` but still tries to add `record` to `good_records`

debboutr:
@kdionisio few things that have come up with this ticket...

1. The `cpcat_code` field was set up to only hold 20 chars, the csv that you are using holds more than that. This draws up an Exception which I have figured out why it doesn't make it bakc to the page so you can see it. This I can fix but if you want to have 22 chars in this field it's always going to be an error.

2. Each of the other Extracted[x] types have used a field `prod_name` which we check must have a 1:1 relationship with the DataDocument, this was looked at in #555.  The ExtracteCPCat has a prod_name field, but it's never being populated from the csv, it would be simple enough to add to the template and then load and check inthe way that we do with the others, but that may not make sense for you guys.

[...]

PR #672, opened 2019-02-13
add upload test fails, fix upload logic, alter cpcat_code field

debboutr:
#666 

this is a punt, there are still 2 outstanding test fails from the `product_dup_check` test, as well as a migration lingering from the `PUC/puc` model update. 

THIS WILL REQUIRE MAKERMIGRATIONS TO BE RUN!!!!! _this will pick up the migration necessary for this tickets needs `ExtractedCPCat_cpcat_code` max_len 20 -> 50 chars.

I have moved to making the `InMemoryUploadedFile` csv come in from a read-in csv instead of writing out the strings in the test, not as easily read as just opening a csv in table format, as well as could reduce the code in other ways with the `make_upload_csv` function that I've added.

This should be open to go in anywhere w/o the migration added to the index.


dashboard/models/extracted_text.py
```diff
@@ -64,8 +64,16 @@ def pull_out_cp(self):
         else:
             return self
 
-
-
+    def one_to_one_check(self, odict):
+        '''
```
mwfrost: Nice documentation of where the method is used.

dmlyons2:
This one looks good to go except it needs a migration to set the hard limit in MySQL for `cpcat_code` to `varchar(50)` @debboutr or @mwfrost can you add that migration file?

mwfrost:
I can add that, but the idea we are experimenting with is for the last reviewer to add the migration file, especially when PRs are all stacked up like this.

Issue #654, opened 2019-02-05
Move ExtractedText foreign key from Extracted[x] child objects into RawChem


mwfrost:
Proposed by Rick in the Slack, this is a good next step after setting up RawChem as a base class in #626 

Every RawChem object has a datadocument, so there is no good reason to store the datadocument foreign key on the child objects rather than on the parent.

mwfrost:
In ticket 627, I added a stopgap `get_data_document()) method to `RawChem` that hunts for the data document subclass by subclass. That will be obsolete once `RawChem` can be directly queried for the ExtractedText key, so it can be removed from the model.

PR #681, opened 2019-02-14
654 move extext

mwfrost:
#654 

This should conclude the recent round of refactoring.


dashboard/models/raw_chem.py
```diff
@@ -1,9 +1,12 @@
 from django.db import models
 from .dsstox_lookup import DSSToxLookup
+from .extracted_text import ExtractedText
 from model_utils.managers import InheritanceManager
 from django.apps import apps
 
 class RawChem(models.Model):
+    extracted_text = models.ForeignKey(ExtractedText, related_name = 'rawchem', 
+        on_delete=models.CASCADE, null=False, blank = False)
```
debboutr: should we still have this in the field, I think we might have just needed it to create the field and then populate, but I don't think there is a need to have a child object that isn't connected to an `ExtractedText`, right?

dashboard/models/extracted_list_presence.py
```diff
@@ -16,30 +14,19 @@ class ExtractedListPresence(CommonInfo, RawChem):
...
         return self.extracted_cpcat.extractedtext_ptr
     
-    @property
-    def extractedtext(self):
-        return self.extracted_cpcat.extractedtext_ptr
-
     @property
     def data_document(self):
-        return self.extracted_cpcat.extractedtext_ptr.data_document
+        return self.rawchem_ptr.extracted_text.data_document
```
debboutr: this is overly picky, but we can write this more directly:
`return self.extracted_text.data_document`

dashboard/models/raw_chem.py
mwfrost: After discussing with Rick, keeping the `null=False` explicit.

Issue #664, opened 2019-02-07
Clean up seed data inconsistencies resulting from 626/627 changes


Daveahern:
The changes to RawChem, ExtractedChemical, etc. have left some inconsistencies in the seed data yaml files.  Remove outdated files and verify contents of new seed files to be used.

PR #693, opened 2019-02-20
664 update fixtures

debboutr:
#664 

this is branched from the head of `654_move_extext_id` so that the fixtures made have extracted text in the `RawChem` model.


debboutr:
this branch is pointless!

Issue #676, opened 2019-02-14
extracted data for new data group type - model


kdionisio:
As a Factotum data curator
I have a need to store extracted data for a new data group type
so that I can integrate a new data type into the Factotum application for ease of priority data record identification

New data group type code = HH

For each data document within a data group type of HH, fields that need to be stored are:
Fields that have a 1:1 relationship with data document...
- filename of the data document (already accounted for in data document model?)
- title of the data document (already accounted for in data document model?)
- date of the data document (already accounted for in data document model?)
- HHE report number (madatory, alphanumeric, max 30 characters)
- study location (not mandatory, alphanumeric, max 50 characters)
- NAICs code (not mandatory)
- Date of sampling (not mandatory, max 75 characters)
- population_gender
- population_age
- population_other
- occupation (max 75 characters)
- facility (max 75 characters)

Fields that will have a many:1 relationship with data document....
- raw cas
- raw chemical name
- media (max 30 characters)
- sampling method (open text)
- analytical method (open text)
- num_measure (alphanumeric, max 50 characters)
- num_nondetect (alphanumeric, max 50 characters)

Definition of Done:
Test that will create the model and check for mandatory fields

Example reports which we are extracting data from:
[HHE_3.pdf](https://github.com/HumanExposure/factotum/files/2861896/HHE_3.pdf)
[HHE_0.pdf](https://github.com/HumanExposure/factotum/files/2861683/HHE_0.pdf)

.csv with example extracted data from the two above reports (this is a bit messy but should help give you an idea)
[dataextraction.zip](https://github.com/HumanExposure/factotum/files/2862019/dataextraction.zip)

mwfrost:
@kdionisio Kathie, is an extraction script being recorded with each one of these HH documents?

kdionisio:
@mwfrost Yes, but it is just a ‘dummy’ script because I think we require an
extraction script to be associated. Data extraction is being done manually
so there isn’t a true extraction script being applied. All HH reports (and
only HH reports) are being associated with an ‘HH report dummy extraction
script’.

kdionisio:
I think I reference the extraction script in the related ticket (677?)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kdionisio:
[...]

PR #698, opened 2019-02-21
676 add extracted HH model

mwfrost:
Ticket #676 adds a new data group type for HHE Reports. This branch includes numerous changes to the seed data to catch up with the current schema.


dashboard/tests/functional/test_datadocument_detail.py
```diff
@@ -32,7 +32,8 @@ def test_absent_extracted_text(self):
 
     def test_script_links(self):
         doc = DataDocument.objects.first()
-        response = self.client.get(f'/datadocument/{doc.pk}/')
+        #response = self.client.get(f'/datadocument/{doc.pk}/')
```
debboutr: there is no need for the line above if we are going to get an explicit URL

dashboard/models/extracted_hhrec.py
```diff
@@ -0,0 +1,40 @@
...
+    sampling_method = models.TextField("Sampling Method", 
+                                        null=True, blank=True)
+    analytical_method = models.TextField("Analytical Method", 
+                                        null=True, blank=True)
+    num_measure = models.CharField("Numeric Measure", max_length=50,
+                                        null=True, blank=True)
+    num_nondetect = models.CharField("Numeric Nondetect", max_length=50,
+                                        null=True, blank=True)                                    
+
+    rawchem_ptr = models.OneToOneField(blank=False, null=False,
```
debboutr: this field does not necessarily need to be expressed in the model def this way, the only thing this changes in the field is the `auto_created` attribute. If we want to be consistent we should either blow all of these defs out of child models OR add the `extractedcpcat_ptr` to ExtractedListPresence

dashboard/models/extracted_hhrec.py
```diff
@@ -0,0 +1,40 @@
+from django.db import models
+
+from model_utils.models import TimeStampedModel
+from .raw_chem import RawChem
+from .extracted_text import ExtractedText
+
+class ExtractedHHRec(TimeStampedModel, RawChem):
```
debboutr: I'm guessing this gives us `created_at` and `updated_at` in the model? If we like how this works are we going to inherit this into other models or continue to use both?

mwfrost: Yes, it would replace our CommonInfo base class.

dashboard/models/extracted_hhrec.py
mwfrost: You are right - if you use the default name for the parent key on a subclass, you don't need to declare anything. I just deleted that explicit declaration from the model, made a migration to remove it, and then ran makemigrations. When it finished up, the column was still in the table where it needed to be.

dashboard/models/extracted_hhrec.py
zachmunzenrider: Might be good to have a ticket to replace CommonInfo across the board. Or would that be too big a PITA?

mwfrost: If we decide to do that, it deserves its own ticket. I don't think it would be too much work.

zachmunzenrider: Yes, definite;ly its own ticket! I'm just wondering if the refactoring is worth it

mwfrost: It occurs to me now that CommonInfo records the user who updates the record - TimeStampedModel does not. That could be a dealbreaker. I should probably roll this back to CommonInfo and redo the seed data and migrations accordingly.

zachmunzenrider: I'd forgotten that! I do think it's worthwhile to capture the user. Maybe we should make CommonInfo a TimeStampedModel ;-)

mwfrost: Looking again at CommonInfo, it does _not_ include the updating user. We are only recording that ad hoc in different forms. What we really need is a TimeAndUserStampedModel abstract base class.

mwfrost: But it's clear that I have just complicated things by using an alternative to CommonInfo and I'll revert to that class.

zachmunzenrider: Cool. It looks like Rick and I are both reviewing this ticket? I'll hold off until you get this checked in and then look again

mwfrost: Thanks. I am going to contain all the changes in the 0090 migration

dashboard/tests/functional/test_datadocument_detail.py
mwfrost: I'd like to clean this up later when I write tests against the data document detail view.

Issue #600, opened 2018-12-03
apply assumed attributes


kdionisio:
For certain PUCs, attributes may be 'assumed' to apply to EVERY product assigned to that PUC. e.g., for the lawn fertilizer PUC, we would assume that the 'outdoor' attribute is assigned to every product within that PUC.

Definition of Done:
1. Way for product owner to assign 'assumed' attributes for a PUC. An assumed attribute will be one or more of the 'associated/allowed' attributes for the PUC. 
2. The assumed attribute will automatically be assigned to every product assigned to that PUC (both for existing data, and for new data).
3. Within the application (e.g. on a given product page), a user should not be allowed to 'de-select' any assumed attribute for a given product.

debboutr:
if there is an attribute assigned to a PUC and we consider that attribute to be `assumed`, meaning it would be automatically assigned w/ a Product and also not able to be unassigned to that Product. Will you ever want that given attribute to be linked to another PUC, but not be assumed? _hope this makes sense_
`a_puc` -> tag('outdoor') -- assumed=True
`another_puc` -> tag('outdoor') -- assumed=False

kdionisio:
@debboutr This situation is absolutely going to happen!

lawn fertilizer PUC --> tag('outdoor') -- assumed = True
paint PUC --> tag('outdoor') -- assumed = False, BUT allowed = True (i.e. outdoor is an 'allowed' attribute for the paint PUC, which a user can select for a given product if it applies)

PR #705, opened 2019-02-26
600 assumed attributes

debboutr:
#600

There is currently only one PUC setup with tags.
[http://127.0.0.1:8000/admin/dashboard/puc/1/change/](http://127.0.0.1:8000/admin/dashboard/puc/1/change/)

if you were to delete a tag that was currently marked as "assumed" you would need to go flag it as assumed in the admin, this is the only place that it can happen currently,
[http://127.0.0.1:8000/admin/dashboard/puctotag/](http://127.0.0.1:8000/admin/dashboard/puctotag/)

once you've checked those out the views to checkout are below:
[product detail](http://127.0.0.1:8000/product/11/) 
[Bulk Assign Products to PUCs](http://127.0.0.1:8000/bulk_product_puc/)
I also added `Products` in the nav bar which may be unnecessary, but is a needed piece of navigation


debboutr:
I also forgot to mention that `prod` has both of the default tables that come with **taggit**. We have written custom tables for tags so whether or not we want to keep those is up for grabs. I don't know if @zachmunzenrider ever tried to use either of them, I thought we might be able to use the `taggit_tag` table like we do the `PUCTag`, but I just stemmed off of what zach had already put in place

templates/product_curation/products.html
```diff
@@ -41,7 +16,36 @@ <h1 class="col-sm-12">
...
+    </tr>
+    {% endfor %}
   </tbody>
 </table>
 {% endblock %}
 
+{% block js %}
+  <script type="text/javascript">
+
+  $(document).ready(function () {
```
dmlyons2: This is a great example why this needs to be moved out of the template and into the js file. You can only have one document ready function per page, if you have more than one you will get an error. Of course js does not inform you of where the error is so you might not even know to look for 2 doc ready functions.

debboutr: this makes sense, but I'm still a little confused about the number of `document.ready()` functions? I see that you've moved the one into the `product.js` file, but where is/are the other/s? this is the list of resources in DevTools in chrome in the product detail view:

image

are one of these other files calling document.ready() as well?

I'm very pink when it comes to all this, so if someone could help to explian, I would be most appreciative!

**ALSO:**
There is some syntax problems w/ your push that is making it fail, seen below, if I delete the strikethrough text below, it works fine.
src="{% static 'js/dashboard/product.js' %} ~~static }}~~"

image

debboutr: I'm also just now noticing that there is already a `js/dashboard/product.js` on dev that is used in the product detail view. This one is used in the product list view, so a different name might be better

dmlyons2: There was only one that I notice but if there is more than one it can cause problems, you do not always get an error as I noted above. It is a problem (more than one doc ready) that has bit me a few times and finding all the javascript that is happening on a given page helps with troubleshooting.

Nothing like pushing with some extra code and breaking working code in an attempt to refactor - my bad. :( 

I am ok with more than one function in the product file as long as it is readable.

Issue #571, opened 2018-11-14
appearance and function edits to PUC and PUC Tags section of product page (5 pts)


kdionisio:
Clean up appearance and functionality of PUC and PUC Attributes sections of product detail page (http://factotum.epa.gov/product/194538/).

Currently, the 'submit' button for the PUC Attributes section of the page is always enabled. Thus, it is hard for the user to tell when their edits to PUC Attributes have been saved.

Definition of Done:
- On the product detail page (http://factotum.epa.gov/product/194538/), please move the 'submit' button in the 'PUC Attributes' section to be on the same line as the 'PUC Attributes' label (but right aligned).
- 'Submit' button in PUC Attributes section should be disabled when no edits to PUC Attributes have been entered which need to be saved. If user selects new PUC Attributes, 'Submit' button is enabled, until user hits 'Submit' button, at which point button becomes disabled again.
- Please rename 'PUC Attributes' label on product detail page to be 'Product Attributes'
- Please remove leading and trailing dashes ('-') from the three levels of the PUC which appear in the PUC box above PUC Attributes

PR #706, opened 2019-02-26
571 puc tag submit

debboutr:
#571

pretty straight forward, this will likely conflict with changes I have in the PR for issue #600 in the `product_detail.html` where I added the hyperlink to `(edit)` the PUC in admin.


dmlyons2:
Note that I pulled the css and js out of the html template and put in static directory.

PR #711, opened 2019-03-01
252 Add raw chem download to Get Data page

mwfrost:
The link downloads a csv file containing all the RawChem records with no link to the DSSToxLookup table. The ID is included, so users can curate chemicals offline and then add them to the database via SQL tools.


mwfrost:
The `tag_submit` test is failing, but that is fixed on a different branch

dashboard/views/get_data.py
```diff
@@ -162,9 +162,19 @@ def upload_dtxsid_csv(request):
...
+
+def download_raw_chems(stats):
+    response = HttpResponse(content_type='text/csv')
+    response['Content-Disposition'] = 'attachment; filename="uncurated_chemicals_%s.csv"' % (datetime.datetime.now().strftime("%Y%m%d"))
+
+    writer = csv.writer(response)
+    writer.writerow(['dashboard_rawchem_id',  'raw_cas', 'raw_chem_name', 'rid'])
+    for rawchem in RawChem.objects.filter(dsstox_id=None):
+        writer.writerow([rawchem.id, rawchem.raw_cas, rawchem.raw_chem_name, rawchem.sid if rawchem.sid else '' ])
+
```
debboutr: it should be `rawchem.rid` not `rawchem.sid` up here right?

dmlyons2: @kdionisio what are you looking for in the column headers for the csv for this ticket? is it `raw_chem_id`, `raw_cas`, `raw_chem_name`, `rid`?

dmlyons2: The ticket says " download a .csv file which includes any raw_cas and raw_chem_names which do not yet have an SID assigned" so I am thinking that sid is correct here @debboutr

mwfrost: No, the file needs to show the RID for the users' benefit.

zachmunzenrider: Maybe we could come up with some _less_ descriptive column names to clear up the confusion! ;-)

mwfrost: Good mnemonic: "R is for 'raw'"

PR #714, opened 2019-03-04
614 test for method of puc classification

zachmunzenrider:


debboutr:
@zachmunzenrider here is a quick bit you can add to the bottom of the `test_bulk_product_puc_post` test that will verify duplicate records can't be made....and won't validate.
```
from django.db.utils import IntegrityError             #this you can import above

duplicate_record = ProductToPUC(product=product,puc=puc,classification_method='MB')
self.assertRaises(IntegrityError, duplicate_record.clean())
```

dmlyons2:
There are ~350 duplicate records in `prod_factotum` that kill this migration. You can see them with this query:
```sql
SELECT p1.puc_id, p1.product_id, p1.classification_method, p1.puc_assigned_usr_id 
FROM dashboard_producttopuc p1
INNER JOIN ( SELECT p2.product_id 
                        FROM dashboard_producttopuc p2 
                        GROUP BY p2.product_id 
                        HAVING COUNT(p2.product_id) > 1) temp
ON p1.product_id = temp.product_id
```
Need to delete the dups prior to pushing this to `prod` with this query:

```sql
DELETE p1 FROM prod_factotum.dashboard_producttopuc p1
INNER JOIN
prod_factotum.dashboard_producttopuc p2
WHERE
p1.id < p2.id AND
p1.product_id = p2.product_id AND
p1.puc_id = p2.puc_id AND
p1.classification_method = p2.classification_method 
```

debboutr:
Makes sense, this improvement removes the possibility for these duplications

Issue #694, opened 2019-02-21
Manually enter extracted text for data group type CP - web form


kdionisio:
As a data curator working in Factotum
I want to be able to manually enter extracted text for documents belonging to data group type chemical presence list (CP)
So that I can efficiently complete and record manual data extraction

Many documents of data group type CP will need to be manually extracted, thus the desire to have a webform or similar available for manual entry of extracted data for documents of group type CP.

Note: similar to #677 and #546 

Definition of Done:
- Ability to manually enter extracted data directly within Factotum application for a data document of group type CP
- fields available in data entry web form should be all user entered extracted text/extracted chemical fields associated with documents of group type CP

debboutr:
@kdionisio what fields would you want to enter in upon the creation of a new ExtractedCPCat object? we have just been displaying in the card...

- doc_date
- cat_code
- description_cpcat
- cpcat_sourcetype

there is also a `cpcat_code` field that we inherit from ExtractedText that is in the fields of the CSV...

kdionisio:
@debboutr I would want the ability to add/fill in the following fields related to this data group type, specifically:

from dashboard_extractedcpcat:
- cat_code
- description_cpcat
- cpcat_code
- cpcat_sourcetype

from rawchem (or whatever raw chemical model is appropriate where this data is stored):
- raw chem name
- raw cas

from extractedtext
 - prod_name
 - doc_date
 - rev_num

Issue #677, opened 2019-02-14
extracted data for new data group type - webform


kdionisio:
As a Factotum user, after creating a data group of data group type HH
I would like the ability to enter extracted data manually for each data document
So that I can enter data into Factotum as reports are being read and extracted individually by a user

Definition of Done:
From the data group index page (http://factotum.epa.gov/datagroup/) for a data group of type HH, I am able to navigate to a form that allows me to enter extracted text fields associated with each data document

User input to the `population_gender` field should be constrained to the choices of "Male," "Female," or "Both." This may involve an edit to the model's field definition.

Test: 
If data document has been uploaded, there is a navigation element to get to the form to enter extracted text and/or if the data document has not yet been uploaded, no navigation element exists.

mwfrost:
@kdionisio Do I understand correctly that a csv file will be uploaded to populate the *data documents* in a data group of type HH, but there is no csv upload to create the *extracted text* equivalent?

kdionisio:
@mwfrost correct. A 'registered records' csv will be uploaded to register each of the data documents (same process/workflow/functionality that already exists). But there is no 'extracted text' csv to be uploaded. The documents will be manually extracted one by one so a webform type functionality for data entry for one data document at a time seems most useful.

mwfrost:
@kdionisio Thanks. This is the first such instance of manually entering extracted data for a document, right? Will there likely be others?

kdionisio:
@mwfrost It is not actually! Habits and Practices data documents are set up to be manually extracted and input via webform.

Yes, it is certainly likely there will be other data documents in the future that will be manually extracted also.

mwfrost:
@kdionisio Great. I will try to build a generalized solution for manual data entry.

mwfrost:
@kdionisio The data document detail view includes some conditional logic that excludes the "Products" card when the document's data group is "Habits and Practices." We now have other data group types that will not link products to the documents, right? This new HH document type, for example?

kdionisio:
@mwfrost
Correct the HH doc type also does not link to products!

debboutr:
@kdionisio what are the fields that you will want to have in the HH child form? I have 

- raw_cas
- raw_chem_name

are there others you want in the card for the HH children?

kdionisio:
Are you referring to the web form data entry page for docs in data group type HH? If so then all of the fields listed in #676 would need to be present.

[...]

PR #716, opened 2019-03-05
694 add extext dd view

debboutr:
#694      , #677 

whole lot going on in this one. this accomplishes the task of allowing the user to add/edit `ExtractedText` records for both `CP` and `HH` group types. This should accomplish what the above tickets are asking for. There may be a clash with the fields in the `ParentForm` used to edit the extext instance in the data document detail view, but we can't tease that apart til we determine what will be used in QA first. The form generating function, `create_detail_formset`, is used for both views.

I fixed the tabbed div layout in the get_data view as well as the accordion in the datagroup/add to find group types. Also, you will see I used a bit of styling in the `datagroups/` list view as well. Probably not the coolest thing we can do, but it's better than nothing.

As much as I want to help out, this is likely the last big ticket that I can do for a while so I can tie up loose ends here before I leave in a few weeks.


dashboard/views/data_document.py
```diff
@@ -95,3 +94,30 @@ def dg_dd_csv_view(request, pk):
...
+                                    data_document_id=pk)
+    form = ParentForm(request.POST, instance=exttext)
+    if form.is_valid():
+        form.save()
+        doc.extracted = True
+        doc.save()
+        return redirect('data_document', pk=doc.pk)
+    else:
+        extext.delete()
+        return HttpResponse("Houston, we have a problem.")
```
mwfrost: Do we need another ticket for the handling of an invalid form?

debboutr: there is no way to send an invalid form here to this view with the current form fields. I just put this here to see if I could be wrong. The jury is till out on that. All of the form fields that need to validate are just `char` fields w/ a length limit which is restricted already in the `<inputs>`. So for now I'd say no, but if we get to a more customize validation in any of those fields then we may need to. I'd be stoked if you could find a way to get to this page using the interface

PR #724, opened 2019-03-06
701 fix edit behavior in QA page

mwfrost:
The editing functionality on the data document QA page broke when I added the DSSToxLookup related fields to the form. I have fixed the error, and added a selenium test.

The pre_save signal was being sent, but it was not being received by the processing method because the dispatching object was a subclass (ExtractedChemical, for example). I have moved that signal-processing logic into signals.py and added potential dispatchers.

The new selenium test that checks whether editing the raw CAS breaks the link to DSSToxLookup only checks an ExtractedText record. The same sort of test should eventually be built for the other subclasses as we work through the editing forms.


mwfrost:
Zach, the form is now displaying the errors if the user-provided values fail to validate.

Issue #591, opened 2018-11-28
remove 'weight fraction type' field from QA page


kdionisio:
Please remove the 'weight fraction type' field from the QA page (http://factotum.epa.gov/qa/extractedtext/155284/). This field does not need to be QA checked since it is not present on the original document, rather it is a categorization made by the user.

Definition of Done:
weight fraction type is not present on QA page

PR #728, opened 2019-03-07
#591 remove weight fraction type from QA view

mwfrost:
I solved this by adding a new `exclude` parameter to the `create_detail_formset()`method. Now when you call that method from a view, you can specify certain fields to be excluded from the rendered form.

### DIFF
--- a/dashboard/forms.py
+++ b/dashboard/forms.py
@@ -316,7 +316,7 @@ def include_clean_comp_data_form(dg):
         return False
 
 
-def create_detail_formset(document, extra=1, can_delete=False, exclude=None):
+def create_detail_formset(document, extra=1, can_delete=False, exclude=[]):
     '''Returns the pair of formsets that will be needed based on group_type.
     .                       ('CO'),('CP'),('FU'),('HP'),('HH')
     Parameters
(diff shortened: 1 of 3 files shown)


dashboard/forms.py
```diff
@@ -305,54 +316,68 @@ def include_clean_comp_data_form(dg):
...
-
+    extracted = hasattr(document, 'extractedtext')
+
+    def make_formset(parent_model, model,
+                     formset=BaseInlineFormSet,
+                     form=forms.ModelForm,
+                     exclude=exclude):
+        if exclude:
+            formset_fields = list(set(model.detail_fields()) - set(exclude))
+        else:
```
debboutr: it'd be best not to use a default in `exclude` that doesn't throw a `TypeError`
```
In [11]: set(None)                                                                                                                  
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-cd95f5c13c52> in <module>
----> 1 set(None)

TypeError: 'NoneType' object is not iterable
```

mwfrost: How did you get to that error? A test?

Issue #633, opened 2018-12-19
method for grouping PUCs


kdionisio:
As a product owner, 
I would like a way to group PUCS into larger categories by 'type' of the PUC (e.g., by each PUC having a type code)
So that I can easily separate different types of PUCs for categorization, or for data analysis.

e.g., our current set of PUCs are all for 'formulations of consumer products'; I will soon be loading another set of PUCs that are all related to articles, another set for building materials, and another set all related to products used in occupational settings. I'd like a way to easily distinguish between these 3-5 sets of PUCs.

Definition of Done:
- Ability to assign each PUC, from within the admin panel, into one of 3-5 'types' of PUCs
- PUC types to create: unknown, formulations, articles, occupational
- PUC type is required for each PUC
- newly created PUCs can have a default PUC type of unknown
- existing PUCs with a 'created_at' date prior to Jan 1, 2019 can be assigned to the 'formulations' PUC type
- existing PUCs with a 'created_at' date after Jan 1, 2019 can be assigned to the 'occupational' PUC type

debboutr:
PUC has a many-to-many relationship w/ Product, so would we want to enforce the `type` of PUC that a Product can be linked to? or are any fair game for linking?

kdionisio:
@debboutr this is a good question, but the answer is probably the latter, any are fair game for linking. Mostly because we typically would not know a priori which 'type' of PUC a product would be assigned to (e.g., some products from Walmart and Home Depot would be assigned to a 'formulation' PUC, some would be assigned to a 'building material' PUC). I'm sure there would be some sources where the products are so specific that we would know the type of PUC the products would be assigned to, but typically we would not.

We would more use this for things like data searching or faceting (i.e., give me all products including chemical A that are formulations vs. products including chemical A that are building materials). It will also be used when we subset data out to create the input files for our models (only weight fractions from formulation PUCs need to go into our exposure models, but building material PUC data may be needed for another model).

PR #739, opened 2019-03-14
633 categorize the categories

debboutr:
#633

the migration here updates the new `kind` field according to date created so, if remade, you'll need to lift that bit into the new one.

I found a way to pass args into a view from the `url` template tag
`"{% url 'download_PUCs' %}?bubbles=True"`
to determine whether or not we want to filter the PUCs for the bubble plot or return **all** for the csv download


mwfrost:
Should the new "Kind" attribute be shown on the PUC admin page? A lot of PUC management happens over in that part of the site.

dashboard/views/dashboard.py
```diff
@@ -102,13 +102,22 @@ def product_with_puc_count_by_month():
 
 
 def download_PUCs(request):
```
mwfrost: Nice docstring!

mwfrost:
This is the only UI change requested in the ticket:
"Ability to assign each PUC, from within the admin panel, into one of 3-5 'types' of PUCs"

I assume future tickets will add the ability to display the PUC Type on the `/pucs/` page and in the admin list page, but since #633 said nothing about this functionality, I will approve the PR.

Issue #529, opened 2018-10-31
refinements to qa check page


kdionisio:
On QA check page (http://factotum.epa.gov/qa/extractedtext/155284/), the 'extracted text info' box is always editable, but edits in this box can only be saved when the 'edit composition records' button is 'on' (so it is irrelevant to be able to edit the fields when edit composition records button is 'off')

Also, in the header box, the Document Type is not able to be modified

Definition of Done:
- Make 'extracted text info' box only editable when 'edit composition records' button is on (this is same functionality as the 'extracted chemical' info, which is only editable when 'edit composition records' button is on.
- 'Save edits' button should only be visible/enabled when the 'edit composition records' button is 'on'
- In the header box, allow user to change the Document Type (however only document types allowed are those which are children of the data group type to which the document belongs)

mwfrost:
An idea for this page, inspired by Rick's work on the data document detail page: 

What if each editable form were made modal? So rather than clicking the button that toggles editing off and on and then going from form to form, then saving edits and seeing what failed to validate, you instead clicked a button that opened a modal form specific to the UI element that you clicked?

mwfrost:
@kdionisio This page has never supported the ability to edit the data document record, so I have to do some up-front planning about how to integrate this new functionality.
Is it okay if the extracted data editing remains separate from the data document editing? I would prefer to channel the data document edits into their own popup form that ask the user to either save or commit before returning to the QA page. Validating the data document edits alongside all the potential extracted data edits would get complicated.

kdionisio:
@mwfrost hmmm ok...and yes I agree with you that to be cleaner in terms of separating true QA (i.e. fields that *need* to be QAed) from just 'edits we are doing during the QA process because it is convenient' it would be good to separate these two things. I am fine with  your proposal of channeling data doc edits into their own pop-up form or similar.

mwfrost:
@kdionisio As with the DataDocument edits, would it also be acceptable to separate edits to the ExtractedText parent object from the edits to the child records?

For example, if a user is performing QA on a set of CPCat list presence records, there would be one button to click that opens the CPCat attributes up for editing (Doc date, Cat Code, Description CPCat, CPCat SourceType), but editing the child records would use the current combination of buttons to make multiple changes and save them.

kdionisio:
@mwfrost Yes that would be fine!

mwfrost:
New requirement:
All form fields should be displayed in a certain order, to be specified by @kdionisio 

[...]

Issue #736, opened 2019-03-13
remove curated chemical fields from the QA page


kdionisio:
As the product owner
I would like the true cas, true chem name, and SID fields removed from the QA page
Because these fields are not checked as part of the QA process

Definition of Done:
true cas, true chem name, and SID fields do not exist on the QA page (http://factotum.epa.gov/qa/extractedtext/264397/)

debboutr:
this issue got handled in issue #529

PR #743, opened 2019-03-14
529 Refine QA page

mwfrost:
I am ready for someone with a few spare minutes to check out my QA page redesign, specified in ticket #529 

Here are the three requirements specified in the ticket:
- Make 'extracted text info' box only editable when 'edit composition records' button is on (this is same functionality as the 'extracted chemical' info, which is only editable when 'edit composition records' button is on. [*NOTE:* after discussion with Kathie, the ability to edit the ExtractedText record was moved into a modal form. The "Edit" button on the detail record formset only applies to those detail records.]
- 'Save edits' button should only be visible/enabled when the 'edit composition records' button is 'on'
- In the header box, allow user to change the Document Type (however only document types allowed are those which are children of the data group type to which the document belongs)

The formset is tricky to display, since the number of forms per child record can vary from 2 in the case of a CPCat document up to 11 in the case of a Composition record. The long text fields in the HHR documents are especially troublesome to lay out.

Every four fields, the template adds a new row. That increment can be changed.

*Composition* example: `/qa/extractedtext/7/`
*ExtractedHHDoc* example: `/qa/extractedtext/354783/`
*List Presence* example: `/qa/extractedtext/254781/`
*Functional Use* example: `/qa/extractedtext/5/`

I have removed the curated fields (True CAS, True Chem Name, and SID) from the QA display.

The formset uses a table for layout, and you will see a row above each record with some dummy code declaring its CSS selector ID. I've used the RawChem id value as a key here, thinking it might help us with testing and styling.


mwfrost:
Please use data document 7 to confirm that the curated fields have been removed from the QA display, as requested in #736

mwfrost:
Styling the error messages deserves a ticket of its own.

zachmunzenrider:
I'm fine with that. Do you mind making the other change?

mwfrost:
The button stye has been changed.

Issue #687, opened 2019-02-20
create confidence level for extraction scripts


kdionisio:
As a product owner,
I want the ability to assign a confidence level to each extraction script
So that I can utilize this categorization of confidence level of the associated extracted text objects in other functionality in the application

Definition of Done:
- Ability to assign, within admin panel, a confidence level to each registered script of type 'EX' (i.e., extraction scripts only, does not apply to download scripts). 
- Confidence level will be represented as an integer from 1-100 inclusive
- When an extraction script is registered, the default confidence level is 1
- Confidence level is a required field for every registered extraction script (can not be null)
- For all extraction scripts that currently exist in the database, set the confidence level to 1

Action item for Kathie: Upon completion of this ticket, Kathie to assign actual confidence level for each extraction script within admin panel

mwfrost:
This looks simple to implement.

One question: is this qualitative confidence index really so fine-grained that you need 100 values on the scale? Is it going to be expressed as a percentage at some point?

debboutr:
if we add this field to the `Script` model both types of scripts, all types would have the field. Would it be okay to just set all scripts w/ the default `confidence` and then likely never it them OR do we want to make a new model (i.e. subclass) to add this field on to only the 'EX' type?? 
```
    TYPE_CHOICES = (('DL', 'download'),
                    ('EX', 'extraction'),
                    ('PC', 'product categorization'),
                    ('DC', 'data cleaning'))
```
It doesn't seem likely that we are going to have thousands of scripts so keeping this new field in the same table along with the others is probably okay?

kdionisio:
I'm fine with the other types of scripts having a field available for confidence level just knowing we will likely never use it, especially if that makes this much easier to implement.

PR #752, opened 2019-03-15
#687 add field, tweak admin, test

debboutr:
I've added to the admin a filter on `script_type` and for the list display I used a function to populate the confidence column for only `EX` Scripts, I like this, but I couldn't get this updated confidence column to be sortable like it would be if they were displayed for all types. Don't know how many scripts there are likely to be, but it seems okay for now


mwfrost:
From the ticket: "Confidence level is a required field for every registered extraction script (can not be null)"

How should we enforce the conditionally required status of the new field? Shouldn't this field be assigned `Null=True` so that the download scripts can have `None` for their confidence levels?

Letting the validator police the form input might be the only way to manage this. We can't enforce `Null=False` at the database level, since all the download scripts are allowed to have `None` values.

mwfrost:
Maybe download scripts have "100" confidence levels?

mwfrost:
Can you export a new version of `04_PUC.yaml` with this new field in it?

debboutr:
@mwfrost I will update the seed data. As for the nullification of the `confidence` field, I'm not sure how important it is that we have the `min/max` validators on this field, but that would be problematic if we wanted to have `null` for any script that isn't `'EX'`. I don like the implementation in the link you've put above, but as long as all of this table is where we are going to put this field it will have to hold _some_ value, whether that be `null` or `1` didn't seem like it would be much of a deal to @kdionisio when talking through this with her. There are certainly other ways to implement it. You've probably already noticed that  at leas in the `admin` panel I filter out the value for other type scripts in `confidence` column so at least visually it looks appropriate there.

mwfrost:
If we can give download scripts a value of 1 or 100 or 0 or whatever without causing problems, then it's fine to leave the field non-nullable.

Issue #670, opened 2019-02-13
custom QA group and QA page for data group type CP


kdionisio:
As the Factotum product owner
I have a need to customize the QA workflow for data documents in data group type CP
So that I can complete QA according to our project QA plan

Definition of Done:
For all data groups of data group type CP, 
- every data document of this data group type will be QA'ed (each data document forms it's own QA group)
- QA group does NOT need to be created for extracted script identified by dashboard_script.id=29
- on the QA page (http://factotum.epa.gov/qa/extractedtext/155284/), a random 10% of the extracted chemical records associated with that data document should be represented, with a max of 30 chemicals represented
- on the QA page (http://factotum.epa.gov/qa/extractedtext/155284/), the raw CAS and the raw chemical name fields are the only fields which need to appear (i.e., the  raw min comp, raw max comp, unity type, ingredient rank, etc etc fields normally represented on this page do not need to appear)

debboutr:
if the DataDocument acts as the QAGroup, when the ExtractedText record is made for the DD it should also instantiate the QAGroup record?

Do we anticipate that CP records will all be extracted by registration through a CSV? Would we potentially add more to a document through the DD detail view?

If we don't consider an extraction script, at what point and on what page would we begin the QA process? If all chem records come in through CSV registration we can create QAGroup upon successful upload, link to QA would be in the DD detail view already, but if not, that's another can of worms...

debboutr:
the way I'm thinking of this being implemented is to add a BooleanField to the `ExtractedListPresence` model and then at the datadocument detail page up in the div w/ the document fields add in a QA button that would either say Begin QA , or, Continue QA. This button would link to a url of it's own that would flag the random (10%/max30) `extractedListPresence` records to True or if the have already been flagged w/ the BooleanField
```
def flag_qa_list_presence(request,pk):
    if not ExtractedListPresence.objects.filter(extracedtext_id=pk, qa_field=True).exists():
        # do the logic to find random 10% or 30 records and set qa_field = True
    return redirect('extracted_text_qa', args=[pk]))
```
Then, the ParentForm would stay the same in the QA view, but the ChilldFormset would have to be given the queryset of children where only the ones w/ the "qa_field" is True. _this field can be better named, but I'm using just to illustrate._

Then the approval updates really all happen in the `ExtractedCPCat` record which inherits `ExtractedText`, so upon approval those fields can be updated.

[...]

PR #753, opened 2019-03-18
670 custom qa group

mwfrost:
Ticket: #670 
This is the initial step toward building our QA workflow for ExtractedCPCat and ExtractedHHDoc records. Upon choosing a data group, it randomly assigns at least thirty child records (ExtractedListPresence, in this case) to be in the QA set and presents only those records in the QA form.


dashboard/views/qa.py
```diff
@@ -1,14 +1,235 @@
...
 
 @login_required()
-def qa_index(request, template_name='qa/qa_index.html'):
+def qa_chemicalpresence(request, pk, template_name='qa/chemical_presence.html'):
+    datagroup = DataGroup.objects.get(pk=pk)
+    if datagroup.group_type.code != 'CP':
+        raise ValidationError('This DataGroup is not of a ChemicalPresence type')
+    extractedcpcats = ExtractedCPCat.objects.filter(data_document__data_group=datagroup)
+    for extractedcpcat in extractedcpcats:
+        prep_cp_for_qa(extractedcpcat)
```
debboutr: I can see how this would work, but it seems like overkill and leaves this view to process this function every time it's called. Adding the flag to th LP object is a one time deal, I would agree that there should probably be a lock put on editing in the DD detail view once QA has begun. The thing to key in on is if any LP objects have a `qa_flag` == True we now know that QA is in effect. We have 3 states of QA Begin/Continue/Complete. In this list view it seems like LP records will get flagged if the user ever goes to this view and there exist LP records for any CPCat record, shouldn't the biggining be deliberate? meaning the user clicks when they want to start and then at the moment that they do LPs get flagged, editing gets locked and they can go look in the QA view at all of the flagged records to approve or not. We can have a view like this, to me it makes more sense to run with just the DG detail and DD detail pages to handle the logic that has to happen. It can be duplicated in this view, but I'm sure the user, when looking at a DataDocument, will want to see what state of QA the doc is in as well as for the DataGroup. Instead of running each CPCat through this function every time this view is hit we could pop the flagging out into it's own view to flag when the user desires to begin and then redirect to the proper view
```
@login_required()
def flag_qa_children(request, pk ):
    ext = get_object_or_404(ExtractedText, pk=pk)
    rawchems = ext.rawchem.all()
    if rawchems.count() > 30:
        rawchems = sorted(rawchems,key=lambda x: random.random())[:30] 
    for chem in rawchems:
        lp = chem.extractedlistpresence
        lp.qa_flag = True
        lp.save()
    return redirect('extracted_text_qa', pk=pk)
```
So, the user would see 

- **Begin QA** button in this table _if_ no LP records have a `qa_flag`==True
- **Continue QA** button if LP records have flag==True and `qa_checked` == False
- **QA Complete** button _disabled_, if `qa_checked` == True

We don't have to use these names for the states we may even want to add more states, but we are going to have to manage these in a way that is clear. This is one way to do that, but as with everything, there are always others.

dashboard/migrations/0096_QA_for_list_presence.py
```diff
@@ -0,0 +1,24 @@
...
+        migrations.AddField(
+            model_name='extractedlistpresence',
+            name='qa_flag',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AlterField(
+            model_name='qagroup',
+            name='extraction_script',
+            field=models.ForeignKey(blank=True, limit_choices_to={'script_type': 'EX'}, null=True, on_delete=django.db.models.deletion.CASCADE, to='dashboard.Script'),
+        ),
```
debboutr: this `AlterField` I think was left from the way that was originally tried when solving this ticket, but is unneeded. I think it should be a requirement for an ExtractionScript to be needed for a QAGroup. Also, This migration doesn't fall in line with the others that have been merged in anyway

dashboard/views/qa.py
```diff
@@ -1,14 +1,235 @@
...
+def prep_cp_for_qa(extractedcpcat):
+    '''
+    Given an ExtractedCPCat object, select a sample of its ExtractedListPresence children
+    for QA review.
+    '''
+
+    from random import shuffle
+    QA_RECORDS_PER_DOCUMENT = 30
+
+    if extractedcpcat.rawchem:
```
debboutr: this will always evaluate to `True`, even if there are no child objects so the `else` will never be used.
```
In [5]: for cp in ExtractedCPCat.objects.all(): 
   ...:     if cp.rawchem: 
   ...:         print(cp.rawchem.all()) 
   ...:                                                                                                                             
<InheritanceQuerySet [<RawChem: hydrofluorocarbon 152a>]>
<InheritanceQuerySet [<RawChem: sd alcohol 40-b (ethanol)>]>
<InheritanceQuerySet []>
<InheritanceQuerySet []>
```

dashboard/views/qa.py
mwfrost: I agree that this process should look more like the Script-based one at `qa/extractionscript/`. The "QA Status" column on `qa/chemicalpresence/{datagroup_id}/` would show one of those three states for each data document. If the link says "Begin QA", then clicking it will run the prep routine on that data document.

dashboard/migrations/0096_QA_for_list_presence.py
mwfrost: It's weird - the `AlterField` operation corresponds to this line of code in the QA Group model, which has not been edited since November:
`https://github.com/HumanExposure/factotum/blob/b49a8427953bf5e25e3937b2d9f93c753977316e/dashboard/models/qa_group.py#L11`

It seems like it should be there. I'm going to put it into its own migration.

debboutr: If it hasn't changed, use
`git checkout <older commit hash> -- dashboard/models/script.py` to reset and it likely won't show up in makemigrations

mwfrost:
@debboutr  What if we condensed the index page that shows data groups and the following page that shows data documents into a single QA index for CPCat records? It would show all the CPCat documents, organized by data group. Once we made time to do more front-end stuff, we could use accordion collapsing to show and hide each group's documents.

Issue #432, opened 2018-09-26
Add another test for DataGroup.get_dg_folder() method


mwfrost:
The get_dg_folder() method was added to the DataGroup model and a functional test was written, but an error could have been detected if the method had been tested on all the data groups in the seed data

mwfrost:
Consider adding an error page that the method delivers when no folder is found

PR #759, opened 2019-03-19
432 Add test that covers every data group in the seed data

mwfrost:
#432


mwfrost:
@zachmunzenrider You are right: It is not a very compelling test.

Issue #657, opened 2019-02-06
ability to upload csv file


kdionisio:
As a Factotum data curator, 
I would like the ability to upload/store/download a .csv file as a data document
So that I can have the .csv file accessible in the future for easier data processing and data extraction (and because I have .csv files that exist as the original data document)

Definition of Done:
- Ability to register, store, and upload .csv files as data documents
- current pdf icons would now become conditional based on file extension (csv or pdf)

kdionisio:
**Note** after testing we have realized the application already allows for registration and upload of .csv (and .txt) document types.

PR #768, opened 2019-03-22
657 csv datadocuments

mwfrost:
Ticket #657 

CSV and HTML files should be allowed as Data Documents. The current system seems to support this, so I have changed the file icon on the data document detail page and added more testing.


mwfrost:
Shortly after I opened this ticket, @kdionisio asked about using HTML files along with csv and pdf. So I decided the generic file symbol is appropriate, since we are going to support all kinds of files.

PR #770, opened 2019-03-25
Unified 529 and 670 PR

mwfrost:
This branch combines 529 and 670 - I reconciled all the merge issues. Once 529 and 670 are both approved, this branch would be the best way to bring their respective changes into the sprint branch.


mwfrost:
Note to the reviewer:
In the course of #529 , I split the template up into separate files. There are now a few different versions of the extractedformset template. I settled on the table-layout version, but I built the others in order to experiment with alternate layouts, and I left them in the branch in case someone rejected the table design. 
`templates/forms/extracted_formset.html`
`templates/forms/extracted_formset_table.html`
`templates/forms/extracted_formset_ul.html`

zachmunzenrider:
I'm happy with this unified branch. Not clear on whether this means I should go ahead and merge it into inchworm-dev myself, so I will not do that!

PR #775, opened 2019-03-27
558 load sids

dmlyons2:


dashboard/views/chemical_curation.py
```diff
@@ -0,0 +1,82 @@
...
+                    except Exception as e:
+                        print(e)
+                        pass
+
+        except Exception as e:
+            messages.error(request, "Unable to upload file. " + repr(e))
+    if records_processed > 0:
+        print(records_processed)
+        data.update({"records_processed": records_processed})
+    print(data)
```
mwfrost: @dmlyons2 Can you remove these `print()` statements? They appear in every test run.

dmlyons2: doneB

debboutr: ha, well this is pretty funny
```
        print('\nprint(some very important test output)')
        print('|￣￣￣￣￣|')
        print('|   HI      |')
        print('|           |' )
        print('|  MATT     |' )
        print('|           |')
        print('|___________|')
        print('(\__/) || ')
        print('(•ㅅ•) || ')
        print('/ 　 づ')
```

zachmunzenrider:
One more thought on this--you might want to use the `csv.DictReader()` like we do starting on line 253 in `views/data_group.py`. It handles a lot of the field-manipulation code that's being done manually in this ticket, and as an extra benefit makes the code more readable.

zachmunzenrider:
After discussion with Dave about the use case for this ticket, we agreed that it should pass review as-is

PR #777, opened 2019-03-28
670 fix: non-ticketed branch

mwfrost:
The new conditional logic (added in #670 ) that renders only the QA-flagged child records under each ExtractedCPCat parent was built to eventually apply to both CPCat and HHE document types. However, the HHRec child table does not yet have the qa_flag attribute, so if you tried to open this HHDoc page:
`http://127.0.0.1:8000/qa/extractedtext/354783/`
you got an error.

The corrected QA view here only tries to exclude the records without `qa_flag = True` after checking to make sure that the model includes that field.


mwfrost:
Closed because it should not have had `dev` as the target

PR #793, opened 2019-04-05
757 manually add extracted text for CO documents

mwfrost:
The same manual entry method that was added for CPCat document has been extended to documents in the CO (Composition) group type. This branch includes a detailed example test for the benefit of new team members unfamiliar with Selenium testing.


dashboard/views/data_document.py
```diff
@@ -16,7 +16,7 @@ def data_document_detail(request, pk):
     template_name = 'data_document/data_document_detail.html'
     doc = get_object_or_404(DataDocument, pk=pk, )
     code = doc.data_group.group_type.code
-    edit = 1 if code in ['CP', 'HH'] else 0
+    edit = 1 if code in ['CP', 'HH', 'CO', ] else 0
```
debboutr: might be worth considering moving the `extractability` of docs in to the model:
```
def extractable(self):
    return self.data_group.group_type.code in ['CP', 'HH', 'CO', ]
```
_this list could also be pulled out into settings and managed there_
`manually_extractable_group_types = ['CP', 'HH', 'CO', ]`, then

```
from settings import manually_extractable_group_types 
...
def extractable(self):
    return self.data_group.group_type.code in manually_extractable_group_types 
```
then, in line 109 of `data_document_detail.html` we can just use the `doc` object to manage instead of having to maintain the extra `edit` object in the view
`{% if edit %}` _goes to..._
`{% if doc.extractable %}`

mwfrost:
The change Rick suggests would be a nice complement to this logic in the test 
`dashboard.tests.functional.test_datadocument_detail.TestDynamicDetailFormsets`
```
if code in ['CO','FU','HP']:
                    error = (f'{model.__module__} should have the same number'
                                                        ' of forms as instances')
                    self.assertEqual(num_forms, children, error)
                # Editable types
                if code in ['CP','HH']:
                    error = (f'{model.__module__} should have one more forms'
                                                                ' than instances')
                    self.assertEqual(num_forms, children + 1, error)
```

The test is using the list of group types to decide whether a data document should be editable (and therefore whether it should have an extra form on the formset for adding a new record). Moving that logic to the model is a good idea

mwfrost:
I created a new `@property` on the data document model that says whether the record can be edited in the detail page. The view and the test both use it now instead of checking a hard-coded list.
The new method still uses a hard-coded list, though.

PR #794, opened 2019-04-08
558 load sids

dmlyons2:


mwfrost:
The branch looks good, but it has several merge conflicts. The ticket branch has not had `dev` merged into it since over a week ago. These conflicts need to be resolved on the ticket branch before it gets merged into `dev`.

Issue #800, opened 2019-04-10
increase length of url field (script model)


kdionisio:
As the product owner
I would like to increase the length of the url field in the script model
so that I can accommodate longer urls 

Currently the url field in the dashboard_script model is max 100 characters.

Definition of Done:
Increase max size of url field in dashboard_script model to 225 characters

PR #806, opened 2019-04-10
#800 change field length for 'url' in Script model, add test

debboutr:
added a unit test for field length verification

## you'll need to `makemigrations` and `migrate` for this branch!


mcovalt:
Not critical, but something cool to think about:

```python
url = [f for f in Script._meta.fields if f.name == 'url'][0]
```

... could be...

```python
url = next((f for f in Script._meta.fields if f.name == 'url'))
```

This has the advantage of not performing a conditional check on every element and not loading the whole list comprehension in memory.

PR #821, opened 2019-04-15
746 ability to edit data document title

### LINKED ISSUE #746: ability to edit data document title
As a factotum user
I would like the ability to edit the data document title from the data document page
So that I can make edits and corrections to the data document title as necessary

http://factotum.epa.gov/datadocument/133494/

Definition of Done:
- ability to edit data document title from data document page
~- ability to edit/change download script from data document page~
(Note it is ok to combine all editing functionality for data document page header into one 'editing' modal/window - just note that the *only* fields in that header which should be editable by the user are the data document title, the data document type, ~the download script~, and the Note)

Updated definition of done 4/11/2019 - strike through added by Kathie, see below comment for explanation

mcovalt:
This adds a data document form clickable via Data Document detail. It removes
1) the unused modal to edit data documents from Extracted Text QA page
2) the redundant modal editor for notes.

Closes #746


mwfrost:
Thank you for cleaning up the proliferation of modal screens. Looks great.

mwfrost:
I like these new test mixins, too.

Issue #786, opened 2019-03-29
Can't edit children twice in QA view


debboutr:
ExtractedChildren can only be edited once with a blank QANotes instance
(Note this is a problem both when editing existing child records and saving >1 time, and when adding >1 new child record on QA page - for the latter, user must enter the new child record, then save edits in order for another blank child record to appear, then enter the second new child record, then save again, causing same error)

**To Reproduce**
1. Go to `qa/extractedtext/<pk>`
2. Edit a child object, then save
3. Edit a child object again, then save

**Expected behavior**
We want to see the QANotes, if added, get carried through the save process when editing child objects

**Screenshots**
![validationQAnote](https://user-images.githubusercontent.com/7052993/55267258-67891d80-523e-11e9-9557-a268dc4684dc.png)

**Additional context**

When a QANote model object is instantiated, which happens upon the GET request in this view, the `qa_notes` field is initially == `None`. The request.POST object that we are instantiating the form with
`notesform = QANotesForm(request.POST, instance=note)`
does not have that field in the form, it is coming specifically from the `extracted_formset_table` form, which doesn't hold the qa_notes input
upon the first save, the `qa_notes` field becomes an empty string `''`, but when saved again this won't pass validation.

Definition of Done:
Ability to edit and save QA form >1 time without seeing an error (including ability to add and save >1 child record without seeing an error)
Note solution should be simplest solution that allows for multiple saves of the QA page without seeing an error, understanding that QA notes will only be saved when the 'approve' button is used, not when the 'save' button is used.

mwfrost:
Something about the control flow in the `qa` view is causing the page to invalidate the document_type attribute after a save. 

The document type shown with a GET request:
![sc_2019417_142645_68.png](https://images.zenhubusercontent.com/5be34a63f724c70c7221c972/9e90f47e-643d-4d75-9782-dbc3d207cff5)

With a POST request:
![sc_2019417_142633_70.png](https://images.zenhubusercontent.com/5be34a63f724c70c7221c972/1e0c553e-8e02-4629-bbcf-0d2b29e0146f)

I will fix this in the course of correcting the other problem.

PR #829, opened 2019-04-19
786 Repeated edits fail on QA page

mwfrost:
Ticket: #786 

The validation process on the Extracted Text QA page was causing errors and it was preventing users from saving more than one round of changes to children of an extracted text object.
In order to resolve this problem, I made several changes to the Extracted Text QA page.
*The Approval Validation Process*
An Extracted Text object can be approved if:
- none of its child records have been edited
OR
- its child records have been edited AND a QA Note has been added

We were previously enforcing this business rule when we validated the QANotes form. That was not the appropriate place to do it. Now the validation process uses a method on the ExtractedText model. The `is_approvable()` method contains the conditional logic, which can be updated later as the approval workflow changes.

*QA Notes Editing*
Previously, a user could not save QA Notes without clicking "Approve." The notes form was where the system checked to see whether the extractedtext object was eligible for approval. This was linked to the problem that inspired this ticket.
I have replaced the old form with an AJAX approach that allows the user to save QA notes at any time in the process without refreshing the entire page. 

*Approval Endpoint*
The approval process was previously being managed in the `qa_extraction_script()` method in `views/qa.py`, based on a conditional fork. I have moved that logic into a separate URL/view endpoint. 

*Warning and Approval Messages*
Rather than using form validation to alert the user to a failed attempt at approval, I have added Messages to the main template. If the user clicks Approve on an extracted text record with modified child records but no QA notes, an error message appears when the page refreshes.


dashboard/forms.py
```diff
@@ -100,7 +100,9 @@ class Meta:
         model = QANotes
         fields = ['qa_notes']
         widgets = {
-            'qa_notes': forms.Textarea,
```
mwfrost: This change adds an id to the QA Notes text area

static/js/dashboard/qa_extracted_text.js
```diff
@@ -0,0 +1,128 @@
...
+            $('.detail-control').removeClass('unlocked');
+            $('.detail-control').prop('disabled', true);
+            //$('.extext-control').removeClass('unlocked');
+            //$('.extext-control').prop('disabled', true);
+            //console.log("Has disabled class been added to controls?");
+            //console.log($('.detail-control').hasClass('disabled'));
+
+        }
+    }
+    // Save notes on submit
```
mwfrost: The javascript code for the AJAX functionality, including the chunk of code to manage the CSRF protection, came from this article:
https://realpython.com/django-and-ajax-form-submissions/

mcovalt:
This passes both the regular and Selenium tests for me.

Issue #637, opened 2018-12-20
restrict data document type codes to be unique


kdionisio:
When entering a new data document type in the admin panel, user specifies a 2 letter 'code' to reference this data document type.

Application currently allows user to enter a two letter code which is already in use (in dashboard_documenttype.code). This seems to be contradictory to the idea that the code is used to uniquely reference the type of data document!

Definition of Done:
Restrict document type code to be unique.

Note it is unclear to me if document type codes need to be unique across all document types, or just need to be unique within each data group type (currently there do exist duplicate document type codes across data group types,, however they are unique within data group type).

mwfrost:
This has led to a condition where many documents have a `document_type_id` foreign key to a record in `dashboard_documenttype` where the code is correct, but the group_type parent of the document_type record does not match the document's own group type.
```
select dd.id, dd.document_type_id, 
dt.id as dt_pk, dt.code, dt.title, dt.group_type_id as doctype_grouptype_id,
gt.id as actual_group_type_id, gt.code as actual_group_type_code, gt.title
 from dashboard_datadocument dd 
left join dashboard_documenttype dt
on dd.document_type_id = dt.id
left join dashboard_datagroup dg 
on dd.data_group_id = dg.id
left join dashboard_grouptype gt
on dg.group_type_id = gt.id
where dt.group_type_id <> dg.group_type_id
;
```

PR #834, opened 2019-04-23
823 ability to store document sub-title

### LINKED ISSUE #823: ability to store document sub-title
As a Factotum data curator 
I would like to store a sub-title for a data document
So that I can store the table name in addition to the document title

Currently, the document title is stored in dashboard_datadocument.title

We also have a need to store a 'sub-title' for each document.

Definition of Done:
1. From the data document page (http://factotum.epa.gov/datadocument/1372072/), ability to enter a data document sub title.
2. data document sub title should appear in the 'header' portion of the page, as the first entry (i.e. before data source)
3. data document sub title is max 250 characters, and is not required (is allowed to be null)

Implement editing functionality to build on #746

mcovalt:
This continues the behavior of setting a `NULL` for string-based fields, but take a look at the [Django docs on this](https://docs.djangoproject.com/en/2.2/ref/models/fields/#null):

> Avoid using `null` on string-based fields such as `CharField` and `TextField`. If a string-based field has `null=True`, that means it has two possible values for “no data”: `NULL`, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not `NULL`.

Closes #823


mwfrost:
Would it be smart to set a default of `""` on this field?

mcovalt:
Ultimately I think so. The reason I chose `default=None` here is because I wanted to continue whatever trend is used in the production database which I assume is string "no data" being `NULL`.

Maybe this is a good PR for a more intrusive commit? Where models are normalized with string "no data" being `""` and `blank=True null=False`?

mwfrost:
The document_type form is causing trouble in the editing screen. 
This is an example from the prod database, where the document_type_id is 1 (Unidentified).

On the data document detail screen:
sc_2019423_121216_67

On the editing screen, the dropdown menu has not been populated with the Unidentified document type:
sc_2019423_121521_734

So when I try to save my edits, the form fails to validate:
sc_2019423_121636_722

Fixing this might be as simple as setting the `initial` attribute for the form.

mwfrost:
Scratch that - initial isn't appropriate here. The problem is that the form is not bound to data when it's rendered. I think it's related to a problem I was having on another ticket

mcovalt:
What is this `pk` so I can see what's going on here? 

The `document_type` choices are being filtered [in the form](https://github.com/HumanExposure/factotum/blob/878c5056f6f4692846f661b14603cec83f9c0f33/dashboard/forms.py#L402-L404) in order to adhere to the [`DataDocument.clean()` validation](https://github.com/HumanExposure/factotum/blob/878c5056f6f4692846f661b14603cec83f9c0f33/dashboard/models/data_document.py#L103-L110). I think this is more a problem with invalid seed data than the way the code works, but I could be wrong.

mwfrost:
I'm using prod data, unfortunately

I haven't yet wiped it out and reloaded the seed data
pick any document in the seed data where the document_type_id is 1 and see if you can reproduce it

mwfrost:
The problem turns out to be in how we are validating the document_type+group_type combination.
A data document with document_type of 1 (Unidentified) is only allowed to belong to a group of group_type 1 (Unidentified).
For a document in group_type 2 (Composition) to be Unidentified, it needs to use the document_type 2 (Unidentified).

The combination of document_type 1 and group_type 2 is not allowed to appear in the document_type dropdown's QuerySet, so it gives up and displays null. ("--------")

The seed data and production data both include documents with document_type 1 and group_type <> 1, so we need to implement a solution that does not break under these conditions. Proposal to allow document_type 1 in the menu's queryset no matter what the group_type is.

mwfrost:
The hotfix works like it should, but the user now sees two rows in the dropdown for "Unidentified". If you add these lines to the end of forms.py, it will change the dropdown labels to include the group type:
```
        self.fields['document_type'].label_from_instance = self.label_from_instance

    @staticmethod
    def label_from_instance(obj):
        return f"{obj.title} (in Group Type: {obj.group_type})"
```

mcovalt:
Repeating here so I can trigger a GitHub link...

Related to #637

PR #847, opened 2019-04-30
836 change cp field labels

justatad44:
Changed CP modal labels to be the same as the data document page.


justatad44:
Yeah, I was wondering about the 'cat_code' vs 'Cat Code'. The ticket seems to be written as making the modal window the same as the data document page, but I felt weird about making the 'Cat Code' in the modal the same as the 'cat_code' in the data document page since that is clearly inconsistent with the other labels. But now I'm a little confused by the sentence-style capitalization comment...are you suggesting that I make the modal label 'cat_code' like the data document page or are you suggesting that I both the modal and the data document page 'Cat code'?

Also, the changes I made to the modal field names were in the extracted_cpcat.py file, not the bs4_form.html file...

mwfrost:
Right, the dynamic form defined in `bs4_form.html ` is smart enough to reflect your changes to the `ExtractedCPCat` model class in `models/extracted_cpcat.py`.

If you change the verbose name of the `cat_code` field to "Cat code" in line 6, the new capitalization will be reflected in the modal form (and anywhere else that dynamically generates the label). But you will still have to manually change that `<dt>cat_code</dt>` label in the template. There should not be anywhere in the interface that displays "cat_code" as a label.

justatad44:
I changed the modal 'Cat Code' to 'Cat code' and changed the data document 'cat_code' to 'Cat code'. I also made a few other labels in the data document consistent with the sentence-style capitalization...

PR #851, opened 2019-05-01
637 unique DocumentType title and code

### LINKED ISSUE #637: restrict data document type codes to be unique

mcovalt:
This changes the schema of `DocumentType`. The FK to `GroupType` is replaced with a ManyToMany relationship so that many `GroupType`s can be associated with a `DocumentType`. This means you can no longer use `DataDocument.document_type.group_type` to identify the `GroupType` of a `DataDocument`. Instead, use `DataDocument.data_group.group_type`. An object manager has been added to `DocumentType` to aid in finding compatible `DocumentType`s for an object related to `GroupType`: `DocumentType.objects.compatible(obj)`.

With this approach to the schema, we introduce a possibility for bad data to remain in the database. If a `GroupType` is removed from the ManyToMany field `DocumentType.group_types`, there may be a `DataDocument` that is no longer valid. To prevent this from occuring, a signal receiver has been added to listen for this event. When this signal has been received, any `DataDocument` that is now invalid will have its `DataDocument.document_type` set to `NULL`. To further prevent this, the ability to delete a `GroupType` from `DocumentType.group_types` is not provided in the admin page.

**This PR includes a data migration!**
See `0107_uq_datadoctypes.py`. This will do a couple things, but the one to take note of is the `update_datadoc_doctype(...)` method. This will preform a bulk update of `DataDocument`s to change their previously non-unique `DataDocument.document_type` to the newly unique `DocumentType`. Proper precautions of such a migration should be taken when migrating in production.

Closes #637


mwfrost:
I like the admin panel functionality:
sc_201952_10508_35

PR #889, opened 2019-05-15
597 boolean fields

### LINKED ISSUE #597: add model properties for BooleanFields where possible
add `@property` to DataDocument model for the `extracted` attribute and similarly to the Script model for the `qa_begun` attribute. 

This will prevent these attrs from being out-of -sync in the DB as well as the seed data.

The boolean fields that we have in the models are:
```
dashboard/models/extracted_text.py:18:    qa_checked = models.BooleanField(default=False, verbose_name="QA approved")
dashboard/models/extracted_text.py:19:    qa_edited = models.BooleanField(default=False, verbose_name="QA edited")
dashboard/models/script.py:27:    qa_begun = models.BooleanField(default=False)
dashboard/models/data_document.py:18:    matched = models.BooleanField(default=False)
dashboard/models/data_document.py:19:    extracted = models.BooleanField(default=False)
dashboard/models/qa_group.py:14:    qa_complete = models.BooleanField(default=False)
```
If any of these could be moved into a model property, it might be worth considering as well.

mwfrost:
Closes #597 (partly)

The `DataDocument.extracted` field was causing us trouble, since it had to be kept in sync with whether or not the document had an ExtractedText related to it. I've removed the field and replaced every instance of it with either `extractedtext__isnull=False` or a new `is_extracted` property method.


mwfrost:
I've removed the offending attempt at setting the `is_extracted` property. The method it's inside should probably go away, as should the entire `habitsandpractices/<int:pk>/` view/template combination. It cannot be reached via the interface and none of the tests touch it.

dashboard/tests/functional/test_datadocument_detail.py
```diff
@@ -90,13 +90,13 @@ def test_add_extracted(self):
...
         doc = DataDocument.objects.get(pk=354784)
-        self.assertFalse(doc.extracted, ("This document is matched "
+        self.assertFalse(doc.is_extracted, ("This document is matched "
                                                     "but not extracted"))
         data = {'hhe_report_number': ['47']}
         response = self.client.post('/extractedtext/edit/354784/', data=data,
                                                             follow=True)
         doc = DataDocument.objects.get(pk=354784)
-        self.assertTrue(doc.extracted, "This document is not extracted ")
+        self.assertTrue(doc.is_extracted, "This document is extracted ")
```
debboutr: not that this is a very big deal, but the message provided here is output if the assertion **fails**, so it should probably stay as it was.

PR #891, opened 2019-05-20
888 replace openiconic with fontawesome

mwfrost:
All the `class="oi oi-widget"` selectors have been replaced with `class="fa fa-widget"`. The static folders have been changed out.


mcovalt:
Wow! 125,323 contributions on a 3 point ticket ;)

The only font issue I found was the `fa-check-square` in the extraction script "QA checked" column e.g. http://localhost:8000/qa/extractionscript/5/.

Looking at the [FontAwesome page for the check-square](https://fontawesome.com/icons/check-square?style=regular) it seems it needs a class telling it whether to be `fas` (solid) or `far` (regular).

Screenshot from 2019-05-24 16-25-25

PR #878, opened 2019-05-13
797 chemical search

mwfrost:
Closes ticket #797 

This branch requires an instance of Elasticsearch running on port 9400, as implemented here:
`https://github.com/HumanExposure/factotum_elastic/pull/1`

The Elasticsearch index must contain the documents for the seed data.

PR #892, opened 2019-05-21
Docker

mcovalt:
This PR allows developers to run two Elasticsearch instances which is necessary to develop on both this repo and the new /HumanExposure/factotum_elastic repo.

There was too much network configuration changes to just comment out the services we don't need. Instead, this implements different folders to house different compositions.


mcovalt:
Closing because changes were included in #878.

Issue #888, opened 2019-05-15
Change icon set


dmlyons2:
## Describe the ideal solution or feature request
We have outgrown our current icon set Open Iconic and need to replace with an icon set that offers more icons. This ticket replaces Open Iconic with the free version of [Font Awesome](https://fontawesome.com). 

## Difficulty, impact, and usage score
| Technical difficulty | User goals | Usage frequency |
|--------------------| --------------------| --------------------|
| small |  Needed for new changes for DD page | daily |

## Who asked for this?
@dmlyons2 

## Definition of Done
Remove Open Iconic and Replace with Font Awesome. 

## What is the test(s)
No test needed.

mwfrost:
Ticket #813 is going to bring in some new Open Iconic references that need to be accounted for on this branch. Please wait until 813 has been merged into dev before approving and moving this.

Issue #813, opened 2019-04-11
creation of a PUC landing page


kdionisio:
As a product owner
I would like to create a PUC landing page for each PUC
So that I have a place to represent up to date summaries for each PUC and can link to the PUC landing page from other pages in Factotum

Definition of Done:
PUC landing page includes summary information for 1 PUC, including:
1. Basic information about the PUC
- PUC name (gen_cat, prod_fam, and prod_type fields from dashboard_puc model)
- definition of the PUC (description field from dashboard_puc model)
- assumed attributes for the PUC
- allowed attributes for the PUC
- 'kind' of the PUC
2. Summary statistics about the PUC
- number of products assigned to that PUC
- number of different curated chemicals (counted by unique DTXSIDs) associated with that PUC
- number of documents associated with the PUC (e.g. a PUC may have 10 products associated with it, but all 10 products may be linked to a single document)

dmlyons2:
![2_813PUCLanding](https://user-images.githubusercontent.com/1336349/57027751-dde4bb00-6c02-11e9-96cd-7ffa20fff3ba.png)

dmlyons2:
![image](https://user-images.githubusercontent.com/1336349/57027789-f48b1200-6c02-11e9-85fb-98b7335fb281.png)

PR #893, opened 2019-05-21
813 puc landing page

zachmunzenrider:
In addition to the landing page itself, I created 2 paths to get to it--one from the PUC List page and one from the Link Product To PUC page. That necessitated adding a Back button to the landing page, so you'll want to test accessing the PUC landing page from both those paths. You can use:

http://127.0.0.1:8000/product_puc/14/
http://127.0.0.1:8000/pucs/

This is more interesting to test with prod data, so maybe try that too if you have it available.

I also updated our fixture PUCs and PUC tags to match prod, and updated tests to match that data.


templates/puc/puc_list.html
```diff
@@ -20,6 +21,9 @@ <h1>
                 <td>{{ puc.prod_fam }}</td>
                 <td>{{ puc.prod_type }}</td>
                 <td>{{ puc.num_products }}</td>
+                <td>&nbsp;<a class="btn btn-success btn-sm" role="button" title="View PUC Detail"
```
mwfrost: This is going to break when it's merged with ticket #888 . The Open Iconic selector will not have anything to point at. If ticket #813 gets merged into dev before #888, then I need to remember to take care of it on #888

dmlyons2: I will fix in the oi ticket when I move it over

Issue #848, opened 2019-05-01
sort ingredients by rank


kdionisio:
As a data curator in Factotum, 
I would like the extracted chemicals to be sorted by rank
So that I can more easily reference the extracted chemicals against the original data document.

Definition of Done:
1. On the QA page (i.e. http://factotum.epa.gov/qa/extractedtext/1365100/), and on the data document page (i.e. http://factotum.epa.gov/datadocument/1341530/),
  - raw chemical records should be listed in ascending order by ingredient rank, if ingredient rank is populated
   - if an extracted chemical record does not have an ingredient rank, those records should appear at the end of the list
Update 5/9/2019 by KLD
 ~ - default ordering when ingredient rank is not populated OR when ingredient rank is not applicable for the data group type is alphabetical by raw chemical name~
- default ordering when ingredient rank is not populated OR when ingredient rank is not applicable for data group type is to leave chemical entries in their default order

mwfrost:
A handy Meta option for this:
https://docs.djangoproject.com/en/2.2/ref/models/options/#order-with-respect-to

PR #900, opened 2019-05-22
#848 sort extracted chemicals by ingredient rank with nulls last

justatad44:
Changed data_document.py and added test to test_datadocument_detail.py.


mcovalt:
There was an issue where this was preforming a query on `ingredient_rank` regardless of whether the `ExtractedChemical` object had that field. @debboutr resolved the issue. I unassigned myself so he could take it.

PR #902, opened 2019-05-23
Correct boolean query

mwfrost:
The "upload extracted chemicals" button was not appearing because my code was incorrectly testing whether all of the data group's records had been extracted yet.


mwfrost:
Thanks! The hotfix SOP is to create another ticket in the standard backlog that includes a test.

Issue #871, opened 2019-05-09
fix uncurated chemicals download


kdionisio:
Currently, the 'Download uncurated chemicals' button on this page: http://factotum.epa.gov/chemical_curation/
gives the following error after running for 5-10 minutes:
Proxy Error
The proxy server received an invalid response from an upstream server.
The proxy server could not handle the request GET /dl_raw_chems/.

Reason: Error reading from remote server

Note that currently this download includes ~500,000 raw chemicals

Definition of Done:
Option for downloading uncurated chemical list which does not time out on server. 
- user should specify data group they wish to download uncurated chemicals from and limit download to 10,000 chemical records

mwfrost:
https://docs.djangoproject.com/en/2.2/howto/outputting-csv/#streaming-large-csv-files

mwfrost:
I have not written a test to check this fix against a data group as large as data group 160 in the prod database (78518 records), but I did test that download in the interface and it worked as specified. Only the first 10,000 records were downloaded. 
![sc_2019529_62836_25.png](https://images.zenhubusercontent.com/5be34a63f724c70c7221c972/bfae08ea-7515-416b-a5f0-3d18f9fb5eb0)

PR #906, opened 2019-05-29
871 fix uncurated chem download

mwfrost:
Closes: #871 

The csv file of uncurated chemicals was taking too long to download, and was sometimes stalling entirely. This branch forces a user to choose a data group to download, and restricts the results to 10,000 records per csv file. 

It would be good if, in a subsequent ticket, we could replace the queryset of the dropdown menu with an aggregation. It currently runs a SQL query for each data group.


mwfrost:
Have you disabled the 10k limit to see what happens? I forgot to try that.

mcovalt:
I just tried it on a set of 78,518 records and it worked well. The only place I can see potential performance improvement is to eliminate the branching [here](https://github.com/HumanExposure/factotum/blob/61a6f10f74bb6ff4b33b90ac1a8c30721765b0c8/dashboard/views/chemical_curation.py#L113), but I bet that's insignificant.

PR #923, opened 2019-06-09
879 data document detail redesign

debboutr:
A lot going on with this one, I added the notes I made as I was going along with this one below. This was one real mutha of a ticket! I spent a lot of time going between FireFox and Chrome trying to get thi8ngs to match up with the scrollspy, it's still not all there in FireFox, but it's working good in Chrome.

- there may be a need to add the ability to link a chemical to a DSSToxSubstance in the card, for now, I left it with a link to the substance if one exists.

- the CP list presence chemicals won't be given tags when **added** at the moment.
- Why don't we use the ExtractedCPCat model to save tags?
- the formset validation and save is back in the detail view to populate the forms w/ errors. the only error that I think can happen from the user is not having the `unit_type` selected in `'CO'` group type

- there is a bit of JS to set the `.scroll_div` to a max height of the available window 

- the QA button is a bit different in the ExtractedText and shows either an empty checkbox or a checked box if QA has been performed

- when edits are done on the chemical cards it should return to the card that you are at.


templates/data_document/chemical_card.html
```diff
@@ -0,0 +1,106 @@
+{% load widget_tweaks %}
+
+<div class="card mt-2 big-bottom" id="chem-{{ form.instance.pk }}">
+    <div class="card-body shadow rounded border border-secondary">
```
mwfrost: Why do the cards have the `secondary` selector? There is enough contrast without it. The additional line just complicates the page.

mwfrost:
Why do we need to put all the Fisher-Price colors on our buttons? Could we use this redesign to get away from this precedent where every "Delete" button is red, etc?

What's the benefit of the heavy-handed color coding, anyway?
sc_2019610_141033_379

templates/data_document/data_document_detail.html
```diff
@@ -4,330 +4,469 @@
...
+        {% endif %}
+    </div>
+    <div class="col-lg-2 d-flex border-bottom">
+        <div class="btn-group my-auto"
+             role="group"
+             aria-label="Data Document Modification">
+            <a class="btn btn-sm btn-info"
+               role="button"
+               title="link to document"
+               href="{{ doc.pdf_url }}">
```
mwfrost: Was it decided not to make this icon conditional on the file format (pdf, csv, txt)?

templates/data_document/data_document_detail.html
```diff
@@ -4,330 +4,469 @@
...
+         id="title">
+        <h1>{{ doc.title|truncatechars:56 }}</h1>
+        {% if doc.subtitle %}
+                <h4 class="text-secondary">
+                    {{ doc.subtitle|truncatechars:90 }}
+                </h4>
+        {% endif %}
+    </div>
+    <div class="col-lg-2 d-flex border-bottom">
+        <div class="btn-group my-auto"
```
mwfrost: The ticket specifies that the document's `url` should appear here if it exists.

templates/data_document/data_document_detail.html
```diff
@@ -4,330 +4,469 @@
...
+                    <div class="d-inline">
+                        <a class="btn btn-sm border-danger btn-secondary"
+                            role="button"
+                            title="Go to QA"
+                            href={% url "extracted_text_qa" extracted_text.pk %}>
+                        <small class="m-0"><b>QA'd</b></small>
+                        <i class="far fa-square"></i></a>
+                    </div>
+                    {% else %}
+                    <div class="d-inline border rounded border-success btn btn-sm btn-secondary"
```
mwfrost: Why does this have a green `success` border?

templates/data_document/data_document_detail.html
```diff
@@ -4,330 +4,469 @@
...
-                <!-- Modal -->
-                <div class="modal fade" id="extextModal" tabindex="-1" role="dialog" aria-labelledby="extextModalLabel"
+                        <span class="fa fa-fs fa-plus"></span>
+                        {% endif %}
+                    </a>
+                </div>
+            </div>
+            {% if doc.is_extracted %}
+                <div class="mt-2">
+                {% for field in edit_text_form.visible_fields %}
```
mwfrost: The `rev_num` field is not being included in the form

templates/data_document/data_document_detail.html
```diff
@@ -4,330 +4,469 @@
...
+    </div>
+    <!-- scrollspy -->
+    <div class="col-lg-2 position-relative bg-white shadow scroll-div">
+        <nav class="pt-2" id="chem-scrollspy">
+            <ul class="nav nav-pills nav-stacked flex-column">
+                {% for form in detail_formset %}
+                {% if form.instance.pk %}
+                <li class="nav-item">
+                    <a class="nav-link" href="#chem-{{ form.instance.pk }}">
+                        {{ form.raw_chem_name.value }}
```
mwfrost: The ticket says the Raw CAS should appear here.

templates/data_document/chemical_card.html
```diff
@@ -0,0 +1,106 @@
...
+                        <dd class="pl-4 mb-0"
+                            id={{ field.id_for_label }}>
+                            {% if field.value %}
+                                {% if field.field.widget.input_type == 'select' %}
+                                    {% render_field field|attr:"disabled" class="form-control col-3" %}
+                                {% else %}
+                                    {{ field.value }}
+                                {% endif %}
+                            {% else %}
+                            <small class="border border-danger rounded pl-3 pr-3">
```
mwfrost: Putting the red `danger` border around every missing value clutters the display.

dashboard/forms/forms.py
```diff
@@ -285,23 +285,10 @@ def __init__(self, *args, **kwargs):
...
-            self.fields['true_chemname'].initial = self.instance.dsstox.true_chemname
-            self.fields['true_chemname'].disabled = True
-            self.fields['SID'] = forms.CharField(max_length=50)
-            self.fields['SID'].initial = self.instance.dsstox.sid
-            self.fields['SID'].disabled = True
 
     class Meta:
         model = ExtractedChemical
-        fields = '__all__'
+        exclude = ('extracted_text',)
```
mwfrost: According to the ticket, the ingredient rank, reported functional use, and weight fraction type fields should display in that order. Explicitly specifying them in the form might be necessary.

templates/data_document/data_document_detail.html
debboutr: this is the icon that was there

templates/data_document/data_document_detail.html
debboutr: the url is in the file button link, I'm not sure the text of the url would do the user much good

templates/data_document/data_document_detail.html
debboutr: I wasn't sure the best way to show the user that the given doc had been QA'd or not in this view, I'll just remove the borde on these for now so the icon is either checked or not

templates/data_document/data_document_detail.html
mwfrost: From the ticket:
> pdf/txt icon (opens data document file, icon should be specific to file type)

This probably should have been its own ticket.

templates/data_document/data_document_detail.html
mwfrost: I think your solution is great, it just doesn't need the green border

templates/data_document/data_document_detail.html
mwfrost: That link is to the source document stored in factotum. The `DataDocument.url` attribute is a link to an external resource, for example:
`http://www.pgproductsafety.com/productsafety/sds/SDS_2015/SDS_June_2015/_VIDAL_SASSOON_PROSERIES_MOISTURE_LOCK_CONDITIONER_99899967_RET_NG_2015020565910.pdf`

templates/data_document/data_document_detail.html
mwfrost: Disregard this requirement for now -- it's getting its own ticket.

templates/data_document/data_document_detail.html
mwfrost: OK, I've been using both CO and CP documents to review the page, since the changes are going to land on both types, but you're right - the ticket says "Redesign layout of data document page for CO group type". I guess someone will discover the missing field in an ExtractedCPCat record

templates/data_document/data_document_detail.html
mwfrost:
```
Menu bar on right side of page; designed as 'quick jump/navigation' menu to bring you to a chemical of choice in the center of the page

Menu bar should include raw name of chemical (dashboard_rawchem.raw_chem_name) as hyperlink
underneath chemical name should see dashboard_rawchem.raw_cas
```

mwfrost:
The changes are working great. There will be some refinements needed to the design, but they will go in another ticket. The non-Composition document types might also need to have some of these changes applied.

PR #933, opened 2019-06-13
848 Fixed issue with ingredient_rank sorting

justatad44:


dashboard/tests/functional/test_datadocument_detail.py
```diff
@@ -112,6 +112,15 @@ def test_delete(self):
...
         self.assertTrue(not doc_exists(), "Document still exists after delete attempt.")
 
+    def test_ingredient_rank(self):
+        response = self.client.get('/datadocument/254643/')
+        html = response.content.decode('utf-8')
+        first_idx = html.index('rawchem-0-ingredient_rank')
+        second_idx = html.index('rawchem-1-ingredient_rank')
+        self.assertTrue(second_idx > first_idx, ('Ingredient rank 1 comes before ' 
+                                        'Ingredient rank 2'))
+
```
debboutr: not to be picky here, but the `rawchem-0-ingredient_rank` text is given as a class and a name when the forms are populated here and will always come out in this order regardless of the chemicals ingredient rank, so to be more explicit with the order coming out right we should maybe key in on the `name` appearing in the order of rank? _hope this makes sense_ This is really close, but I think it'll pas w/o the update on the sorting of chems too

mwfrost: Maybe the test should assert that the value in `rawchem-0-ingredient_rank` is lower than the value in `rawchem-1-ingredient_rank`?

justatad44: Makes sense. How about:
        first_idx = html.index('name="rawchem-0-ingredient_rank" value="1"')
        second_idx = html.index('name="rawchem-1-ingredient_rank" value="2"')

mwfrost: So if the "2" is in the `rawchem-0-ingredient_rank` element, then it will give an error?

justatad44: Yes, it will say that the substring is not found.

PR #934, opened 2019-06-13
905 hotfix test 901

debboutr:
this won't pass with what is currently on `dev`, but when the hotfix commit on `master` gets merged in, it will! I have a strong urge to change the way the `include_extract_form()` function is implemented that I went through with @zachmunzenrider on Monday, but that will likely come in another ticket.


zachmunzenrider:
I updated this branch with the actual bug that Matt fixed in the original 901 ticket, and also an additional fix to another test that was failing. It's also been brought into line with dev and master

Issue #897, opened 2019-05-22
tool tip display for attribute definition on PUC page


kdionisio:
As a Factotum user,
I want to easily be able to view the definition of each attribute on the PUC landing page
So that I can very efficiently understand what the meaning of the attribute is without having to navigate to a different page

Definition of Done:
For the PUC landing page created in #813 (http://factotum.epa.gov/puc/20/), add tool tips which display the definition of the attributes (both for allowed and assumed attributes)

PR #949, opened 2019-06-17
#897 add title to PUC attribute buttons

debboutr:
pretty straight forward, added a couple more defintions to puctags here.

I did notice the test I added to had been using  a truth assertion 
```
self.assertTrue(response_html.xpath('string(//*[@id="assumed_attributes"][count(button) = 2])')
```
this returns: 
```
        
            Assumed Attributes
        
            
                
                    gel
                
                    paste
                
            
        
```
which will evaluate to true, the problem is, it's just a string, and any string, even `"_"` a blank space, will evaluate to true.


debboutr:
Zero thinking involved with merging the yaml! I just added a couple `definition` s to puctags

mwfrost:
The problem I have in mind is the cumulative effect of the yaml changes on this ticket and your other one, PR #941 . I merged the 898 branch into 897 (with "Accept both changes" as my merge reconciliation method) and the tests pass, so I'm going to assume that you made the same changes to the `definition` attributes in `dashboard/fixtures/13_puc_tag.yaml` on both tickets. That would mean that when Dave L merges these into `dev`, he should be able to resolve the conflict with whatever option he chooses. 

Reformatting the yaml for legibility is cool, but, like linting other source code, it destroys the ability to track and inspect other differences in the contents. So if you want to change the content of the seed data by dumping out a new (and differently-formatted) file, please dump the unchanged file in its own commit, then change the content, then dump out the changed version so that there's one commit showing the full-file format change and another commit showing the content changes.

PR #950, opened 2019-06-18
867 chemical landing page bubble plot

### LINKED ISSUE #867: chemical landing page - bubble plot
As a product owner
I would like to add a bubble plot to the chemical landing page
To visually represent information on PUCs containing a certain chemical

Definition of Done:
Add a figure (bubble plot) representing all PUCs which contain the chemical of interest to the chemical landing page created in #815.
- each bubble represents a PUC
- Size of the bubble represents number of products assigned to that PUC AND containing that chemical (chemical presence in a product will be determined only by the DTXSID, we will not consider raw chemical name/raw CAS)
- Hover over each bubble should show PUC name and number of products
- bubble plot should use same color scheme and format as that of bubble plot on main Factotum 'dashboard' page, including three level approach
(Note this is essentially the same plot that currently exists on the Dashboard page, but making that plot conditional on representing only PUCs and products containing a single DTXSID)

mcovalt:
Closes #867 

This also reduces the time to load bubble plots from ~2 seconds to less than 0.01 seconds.


kdionisio:
Yaaaaayyyyy to the reduced load time!!

PR #951, opened 2019-06-18
911 datagroup pagination

zachmunzenrider:
This turned into a bigger ticket than anticipated, in large part because of performance issues. I recommend testing this against a large datagroup, such as the prod data datagroup 49. which has over 11k related datadocuments. The dev version of this page was almost unusable. It still isn't perfect, but I've got it down to 22 queries instead of 1000s. The previous django pagination this replaces generated a whole new request with each page, whereas the new DataTables pagination has it pre-loaded and is very zippy at that point. If at some point folks decide that it's still too clunky, we could move to server-side pagination. At that point the work I've done with this ticket should make that much easier to implement.


dashboard/tests/functional/test_datagroup_detail.py
```diff
@@ -207,12 +207,12 @@ def test_download_raw_comp_data(self):
             ).order_by().values_list('data_group_id',flat=True).distinct()
 
         for dg_id in dg_ids:
-            #resp = self.client.get(f'/datagroup/%s/' % dg_id)
-            resp = self.client.get(f'/datagroup/raw_extracted_records/%s/' % dg_id)
+            print(f'/datagroup/%s/download_raw_extracted_records/' % dg_id)
```
mcovalt:
```suggestion
```
This is a bit distracting during tests. I suggest we remove this print statement.

dashboard/views/data_document.py
```diff
@@ -2,14 +2,10 @@
...
-from djqscsv import render_to_csv_response
 from django.contrib import messages
 
-from dashboard.forms import *
-from dashboard.forms import ExtractedListPresenceTagForm
-# if this goes to 0, tests will fail because of what num form we search for
-from factotum.settings import EXTRA
-from dashboard.models import *
+from dashboard.forms import ExtractedListPresenceTagForm, create_detail_formset, DataDocumentForm, DocumentTypeForm
+from dashboard.models import DataDocument, ExtractedText
```
mcovalt:
```suggestion
from dashboard.models import DataDocument, ExtractedText, Script
```
We use the `Script` model on line 148 in `extracted_text_edit(request, pk)`. Tests fail if not imported.

PR #959, opened 2019-06-19
945 puc prod type bug

zachmunzenrider:


zachmunzenrider:
Sorry for the messy commit. It's really just the one file that's changed!

PR #972, opened 2019-06-20
915 cp qa url bug

zachmunzenrider:
See my ticket comment for some additional details on what I did for this ticket


mwfrost:
Fixed seed data URLs; fixed tests with hard-coded column indexes in XPATH.

Issue #939, opened 2019-06-13
extracted chemicals statistic incorrect


kdionisio:
The 'Extracted Chemicals' statistic on the main Factotum dashboard page (http://factotum.epa.gov/) is not correct.

On production, the value currently displayed for 'Extracted Chemicals' is the count of records in the dashboard_extractedchemicals model

Definition of done:
Extracted Chemicals statistic on dashboard page should display the count of total records in the dashboard_rawchem model

PR #975, opened 2019-06-24
#939 Fix dashboard chemical count

mwfrost:
The dashboard was only displaying the count of ExtractedChemical records. It's now switched to RawChem records, which also include the CPCat records.


dashboard/tests/functional/test_dashboard.py
```diff
@@ -131,6 +131,11 @@ def test_producttopuc_counts(self):
         self.assertIn('Products Linked To PUC', response,
                       'Where is the Products Linked to PUC card???')
         response_html = html.fromstring(response)
+
+        chem_count = response_html.xpath(
+            '/html/body/div[1]/div[2]/div[3]/div/div')[0].text
```
mcovalt:
```suggestion
            '//div[@class="card-body" and contains(h3, "Extracted Chemicals")]/div')[0].text
```
I've been playing around with XPath lately. This locator is a bit easier to understand and more robust if the layout changes.

PR #978, opened 2019-06-27
563 bulk create product number fix

### LINKED ISSUE #563: Bulk Create Products Button - Numbers Off
**Describe the bug**
DataGroup Detail Page - Bulk Create Products is showing a minus 310000 products to bulk create. There are not that many data docs in the Group

**To Reproduce**
1. Go to http://factotum.epa.gov/datagroup/24/
2. See error on mid page right side (Bulk create products

**Expected behavior**
It should show a number that is = to the data documents that do not have products associated

**Screenshots**

mcovalt:
Closes #563

Previously we were just fetching the first product related to a data document in our main query. This caused issues when a data document had many products.

This PR makes the main query include all products.


mcovalt:
> I'm curious about the copious line-break styling in the test. Is this something Black is imposing on us, or something you suggest we do?

Just a style thing. I picked it up when doing [lodash chain methods](https://lodash.com/docs#chain). I think Black will do something similar if it won't fit in the line length.

Issue #883, opened 2019-05-14
tool tips for icons on data doc page


kdionisio:
As a product owner
I want to add tool tips for all linking icons on the data document page
So that a user can hover and see what the icon represents

Definition of Done:
On data document page,
- when hovering over icon linking to dashboard_datadocument.url field value in upper right, tool tip should state: 'Online version of document'
- when hovering over icon linking to dashboard_datagroup.download_script_id linked to dashboard_script.url in upper right, tool tip should state: 'Download script'
- when hovering over icon linking to the actual data document stored on the file server in upper right, tool tip should state 'View data document'
- when hovering over icon in 'Extracted Text' section linking to dashboard_extractedtext.extraction_script_id linked to dashboard_script.url, tool tip should state 'Extraction script'

debboutr:
@kdionisio how are bullets 1 & 3 different? aren't they both the same button? 


![dddetail](https://user-images.githubusercontent.com/7052993/60368747-a8b8d700-99bf-11e9-90e4-a930c9d26ca2.png)

Also,

kdionisio:
Hmm ok, Bullet 3 is correct. Ignore bullet 1. I guess we don't actually
have a place where the field dashboard_datadocument.url is linked at the
moment. I'll put that in another ticket.

PR #984, opened 2019-07-01
#883 update tooltips

debboutr:


templates/data_document/data_document_detail.html
```diff
@@ -28,7 +28,7 @@ <h4 class="text-secondary">
              aria-label="Data Document Modification">
             <a class="btn btn-sm btn-info"
                role="button"
-               title="link to document"
+               title="Online version of document"
```
mwfrost: This is the `doc.pdf_url` link - it goes to the locally archived pdf, not the remote online version. I think Kathie had bullet 1 ( 'Online version of document') in mind when she said "I guess we don't actually have a place where the field dashboard_datadocument.url is linked at the moment."

The link to open the pdf (or whatever format the file was stored as) should be the one that says "View data document."

mwfrost:
```suggestion
               title="View data document"
```

PR #985, opened 2019-07-01
931 list presence keyword page

zachmunzenrider:
NB - an extra "tags" field crept into the ExtractedListPresence model at some point. I determined that I could delete the first of them (which I guess just gets overridden by the second one) with no ill effects, so I did.


dashboard/views/data_document.py
```diff
@@ -133,14 +132,20 @@ def extracted_text_edit(request, pk):
     model = ParentForm.Meta.model
     script = Script.objects.get(title__icontains='Manual (dummy)', script_type='EX')
     try:
-        exttext = model.objects.get_subclass(data_document_id=pk)
+        extracted_text = model.objects.get_subclass(data_document_id=pk)
```
mwfrost: Good cleanup of variable names here

PR #992, opened 2019-07-04
Disabled data group type change in the edit page

justatad44:
Was already disabled for group types with extended docs -> made it disabled for everything


justatad44:
Admin editing of group type for data groups has now been locked down.

Issue #539, opened 2018-11-01
unit type null when no composition data


kdionisio:
Currently, unit_type is a required field when uploading extracted text data for data group type composition. However, this results in records where unit_type is populated but raw_min_comp, raw_max_comp, and raw_central_comp are all null.

Definition of Done:
- unit_type should only be a required field when raw_min_comp, raw_max_comp, or raw_central_comp is populated
- if all three above named fields are null, then unit_type should be null

mwfrost:
@kdionisio We have the "Unknown" unit type. Is that a separate thing from this absence of a unit type because there is no composition data? Or should the unit type be assigned as "Unknown" when a record is imported with no raw_min_comp, etc.?

kdionisio:
@mwfrost You are correct. Unit type should be null if raw_min_comp, raw_max_comp, and raw_central comp are ALL null. Unit type would be required/populated (either with 'unknown' or with another value) if one of the three comp fields is populated. Thus if one of the three comp fields are populated, user should be required to enter a unit type, but one of their options for unit type is 'unknown'

PR #994, opened 2019-07-04
539 unit_type nullable

mwfrost:
Closes: #539 

The `unit_type` attribute on the `ExtractedChemical` model was previously non-nullable with a default of "Undefined." Now it is allowed to be empty, but only if the composition amounts are also empty. This conditional enforcement is handled in the model's `clean()` method, and the QA editing interface now includes error messages.

I also had to change the QA page so that attempting to save an invalid record re-opened the page in editing mode.


mwfrost:
I think all the merge conflicts are instances where I have linted files that have not been linted on `dev`

Issue #1027, opened 2019-07-24
New Search Landing Page (Products, Documents)


dmlyons2:
Frontend replacement of haystack faceted search within application. Desire ability to search on Products by title, long and short description, brand, manufacture and Data Documents by title with the following two facets

Data documents, by type of data document (e.g. no data document, MSDS, SDS, ingredient list, etc.)
Product category, by general category (e.g., no product category, list of ~12 general categories)
Search bar appears on far right side of the navigation bar, on every page of the application. User enters a product title in the search bar. User is then taken to a landing page with search results on product title, with the two facets visible on the left side of the page.

- Search results will be the top 25 displayed with pagination for the remaining
- Search results with state how many results there are and how long it took to search

PR #1041, opened 2019-08-14
645 replace haystack

### LINKED ISSUE #645: Replace Haystack search indexes
The current search functionality uses a single index called `product-index`. The details of the index can be seen here:
`http://localhost:9200/product-index/?v`

This single index could initially be replaced by a new single index, especially since we display the results together, but eventually, we should create different indices that suit different data models. 

The code that preps and loads each object for indexing is here: `dashboard/search_indexes.py`

There are some unused methods in the DSSToxSubstance and ExtractedChemical models that will be reactivated when we switch to the elasticsearch-dsl library. The general structure will be for the `dashboard/search_indexes.py` file to contain python-style object descriptions for each index, while the models will have `indexing(self)` methods to prepare the data for the index. The current Haystack approach puts the data prep into the model-specific index class in `dashboard/search_indexes.py`.

DoD:
- This is a search of products (title, brand_name[aggregation], short_description, long_description, manufacturer[aggregation])and data documents (title)and dd type (aggregation)
- Creation of an index that would represent these results
- Creation of a logstash file that would 'harvest' resultsets daily at 10 pm local to the machine it is harvested on
- Appropriate docker file modifications as needed
**Does not include template layer**

mcovalt:
Closes #645 and #1027 

Depends on https://github.com/HumanExposure/factotum_elastic/pull/6


dashboard/urls.py
```diff
@@ -235,6 +234,7 @@
         "chemical/<int:doc>/create/", views.ChemCreateView.as_view(), name="chem-create"
     ),
     path("chemical/<pk>/edit/", views.ChemUpdateView.as_view(), name="chem-update"),
+    path("search/<str:model>/", views.search_model, name="search-model"),
```
mwfrost: I like the dynamic URL. This will save us a lot of hassle.

PR #1042, opened 2019-08-19
1005 delete extracted text

zachmunzenrider:
I've got dev (with the search changes) merged in and everything seems to be working smoothly. I've introduced a module called bootstrap-confirmation to handle the confirmation piece of the delete. Seems to work pretty smoothly, and if folks agree I think we can add it easily to more places in the app.

Note that after you delete the extracted text associated with an extraction script, that script disappears from the grid. The script still exists, but the grid is filtered for scripts that have associated extracted text, and suddenly the affected script doesn't have that.


debboutr:
Zach,

Could you use the garbage can icon and then the hover class that Matt set up in `factotum.css`. We don't likely need the redundancy of all the text here
image
I'd say you can put the text into the tooltip of the icon and then remove the header title of that column.

debboutr:
Another question that I thought of here that I forgot to mention is, should we be displaying the `Manual (dummy)` script in the table we have in this `qa/extractionscript/` view? if we decide that we do, I would think that we might not want to give the user the ability to delete that one because that data can't be reloaded in the same way..??

zachmunzenrider:
I did the button text that way because the ticket specified it. I think a trash-can icon could be misinterpreted as a way to delete the script itself, rather than the associated text. Do you agree?

I do agree that the `Manual (dummy)` script should be excluded, and I'll change that. Good catch!

kdionisio:
@zachmunzenrider @debboutr I think the trash can icon is good - since the column header is 'delete extracted text' I don't think we need the text replicated in the button. And yes, definitely don't display the manual (dummy) script in this display, good catch!!

PR #1074, opened 2019-08-28
fixing chemical name tooltip

zachmunzenrider:


templates/data_document/data_document_detail.html
```diff
@@ -344,7 +344,7 @@ <h5 class="modal-title" id="extextModalLabel">
                         <p class="m-0"
                            data-toggle="tooltip"
                            data-placement="left"
-                           title="{{ chemical.rendered_chemname|truncatechars:18 }}">
+                           title="{{ chemical.rendered_chemname }}">
```
mcovalt:
```suggestion
                           title="{{ chemical.rendered_chemname }}"
                           data-boundary="viewport">
```

As is, long chemical names start flashing as the tooltip tries to fit inside the nav bar.

This `data-boundary` attribute changes where Bootstrap tries to fit the tooltip to the full viewport. It also properly places the tooltip to the left rather than on top of the text.

Issue #1014, opened 2019-07-11
convert main bubble plot to nested bubble plot


kdionisio:
As the Factotum product owner
I would like to convert the bubble plot on the dashboard page to a nested bubble plot
So that the PUC-product count data is appropriately represented

In the current bubble plot on the Dashboard page (http://factotum.epa.gov/), products are being double counted/doubly represented due to the representation of the bubble plot.

Definition of Done:
Convert bubble plot on dashboard page to a nested bubble plot such as: https://observablehq.com/@d3/zoomable-circle-packing
  - size of bubbles should be represented by number of products in the PUC
  - zoomable feature is requested
  - maintain functionality to click on a bubble and navigate to the appropriate PUC landing page (e.g. http://factotum.epa.gov/puc/162/)
  - color scheme to be used (to be identified by Kathie)
*Note that this plot will consistently represent about 300 PUCs. While number of PUCs (i.e. bubbles) will remain relatively constant, the number of products per PUC (i.e. the size of each bubble) will be constantly increasing.

PR #1087, opened 2019-09-06
1014 Nested Bubble Plot

mwfrost:
Closes: #1014 

*This branch includes `1080_hotfix_into_dev`*

This uses version 5 of the D3js library, while the non-nested bubble plot (still used on the chemical detail page) uses version 4, and NVD3 uses version 3. So the index page is loading three versions of D3. 

We can streamline this with two future tickets: one to replace the chart on the chemical detail page with a chemical-specific version of this nested chart, and another to replace the NVD3 graphs with native D3. 

There are still three search-related tests failing on the seed data:
```
======================================================================
FAIL: test_facets (dashboard.tests.functional.test_search.TestSearch)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/mfrost/Code/factotum/dashboard/tests/functional/test_search.py", line 92, in test_facets
    self.assertIn(expected_total, total_took)
AssertionError: '2 products returned' not found in '3 products returned in 0.008 seconds.'

======================================================================
FAIL: test_input (dashboard.tests.functional.test_search.TestSearch)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/mfrost/Code/factotum/dashboard/tests/functional/test_search.py", line 100, in test_input
    self.assertIn(expected_total, total_took)
AssertionError: '120 products returned' not found in '121 products returned in 0.016 seconds.'

======================================================================
FAIL: test_number_returned (dashboard.tests.functional.test_search.TestSearch)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/mfrost/Code/factotum/dashboard/tests/functional/test_search.py", line 72, in test_number_returned
    self.assertIn(expected_total, total_took)
AssertionError: '33 products' not found in '38 products returned in 0.008 seconds.'
```
These are not related to the ticket.


debboutr:
@mwfrost I just tried using the `d3v5` library with the old bubble plot on the chemical detail page and it seems to work fine, is there another reason we want to keep another v. of d3?? Also, the plot size on the index page comes out bigger than a full screen, shouldn't we try to fit it into a space where we can see it all w/o having to scroll?

PR #1115, opened 2019-10-01
659 validate dg

debboutr:
This ticket moves all of the logic that we had in the view into a formset where the CSV with datadocuments are stored with their associated IDs in the media folder under the DataGroup's UUID and will also create the zip archive which will be used when uploading the documents. There are quite a few errors that I have produced and that the tests check for, here is the group of errors that I made in CSV's

- duplicate filename's
- wrong column header's 
- incorrect URL in the DataGroup AND in the DataDocument
- missing filename
- incorrect compatibility of GroupType to DocumentType

This keeps our registering records process the same, but with the use of formsets and the CSV serializer as opposed to the csv.DictReader that we were using before. If any error is found the datagroup should be deleted.


dashboard/models/data_document.py
```diff
@@ -6,6 +6,10 @@
 from django.core.validators import URLValidator
 
 
+def get_type():
```
zachmunzenrider: Would you name this something more specific? get_type() doesn't really reflect what this is doing. Also not clear on whether this is in the right place. Why is it in data_document.py, and why is it outside of the class def?

zachmunzenrider: I guess maybe it needs to be in this spot so it can be used in the field definition? Even so, I think the name should be more descriptive

debboutr: I was just using the example from the docs
`https://docs.djangoproject.com/en/2.2/ref/models/fields/#arguments`, I'll change the name to `get_default_document_type` unless you have a better suggestion.

debboutr:
@zachmunzenrider I just fixed the function name and the error message for the CSV headers, you still should have been able to load a CSV in if it was correct, try using the `sample_files/register_records_matching.csv` that one works for me.

dashboard/models/data_document.py
zachmunzenrider: That sounds perfect

dashboard/forms/data_group.py
```diff
@@ -273,3 +284,92 @@ def save(self):
...
+    def clean(self, *args, **kwargs):
+        """Errors raised here are exposed in non_form_errors() and will be rendered
+        before getting to and form-specific errors.
+        """
+        with open(self.dg.csv.path) as f:
+            reader = csv.reader(f)
+            header = next(reader)
+        header_cols = ["filename", "title", "document_type", "url", "organization"]
+        if header != header_cols:
+            raise forms.ValidationError(f"CSV column titles should be {header_cols}")
```
mcovalt: The first row of the CSV is read by the serializer and can be accessed from `self.bulk.fieldnames`.

dashboard/forms/data_group.py
```diff
@@ -273,3 +284,92 @@ def save(self):
...
+
+    def save(self):
+        with transaction.atomic():
+            new_docs = []
+            for f in self.forms:
+                obj = DataDocument(**f.cleaned_data)
+                new_docs.append(obj)
+            now = datetime.now()
+            DataDocument.objects.bulk_create(new_docs)
+        made = DataDocument.objects.filter(created_at__gte=now, data_group=self.dg)
```
mcovalt: This seems a little brittle. The time between when `datetime.now()` executes and the `made` variable is created we have a bit of time sending lots of data to the database. This means we have to `.filter(created_at__gte=now, ...)`.

If we manually set `DataDocument(created_at=now)`, then we could lookup the documents at a much higher precision (`.filter(created_at=now, ...)`).

PR #1124, opened 2019-10-03
Synonym tokenizer: factotum

mwfrost:
Closes: https://github.com/HumanExposure/factotum_elastic/issues/5

This branch only works along with `5_synonyms` on `factotum_elastic`. Instructions for testing it are on that PR: https://github.com/HumanExposure/factotum_elastic/pull/16


docker/dockerfiles/Dockerfile.factotum_elastic
```diff
@@ -0,0 +1,7 @@
+FROM elasticsearch:7.3.1
+
+ARG BRANCH
+
+# Put the synonyms file into the container and assign read permissions
+ADD --chown=elasticsearch:root https://raw.githubusercontent.com/HumanExposure/factotum_elastic/"${BRANCH}"/chemical_synonyms.txt /usr/share/elasticsearch/config/chemical_synonyms.txt
```
mcovalt: I don't think `ADD --chown=...` works with the version of Docker on prod. Best bet is to just do a `COPY ...` and the run `chown` in the `RUN` command below.

docker/compositions/elk/docker-compose.yaml
```diff
@@ -69,5 +70,11 @@ services:
       - type: volume
         source: lsdata
         target: /usr/share/logstash/data
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://elasticsearch"]
+      interval: 30s
+      timeout: 5s
+      retries: 3
+      start_period: 10s
```
mcovalt: I was imagining the healthcheck being on the Elasticsearch container. Since the Logstash container is dependent on that container, I was hoping the Logstash service would wait until Elasticsearch reported it was healthy before starting.

In addition, we ought to make sure this works on the production Docker version. The docs state this is new in Docker 1.12 which is quite a few years old. Fingers crossed.

docker/dockerfiles/Dockerfile.factotum_elastic
mcovalt: I just checked and confirmed the `ADD --chown=...` does not work with the Docker version on prod. We'll have to do it manually in the `RUN` command.

docker/compositions/elk/docker-compose.yaml
mwfrost: How about we just remove the healthcheck if there's any question of version compatibility?

mcovalt: Sounds good. I don't think the way I was envisioning it working will work for us. See https://github.com/docker/compose/issues/4369

I think the most robust way to accomplish this is to stick the following script at the start of Logstash: https://github.com/vishnubob/wait-for-it

mwfrost: This is ready for re-review

PR #1137, opened 2019-10-11
keep logstash container running

fang-yan-csra:
Changed the logstash server restart from "no" to unless-stopped, as the logstash container has scheduled nightly job now. Also added DELETE_INDEX_SCHEDULE and REINDEX_SCHEDULE environmental variables to customized the scheduled time. Defaults to 1:25AM and 1:30AM daily.


mcovalt:
We're going to have to think about how we'll implement this in our development environments. There needs to be some way for us to load the fixtures data on-demand.

fang-yan-csra:
> We're going to have to think about how we'll implement this in our development environments. There needs to be some way for us to load the fixtures data on-demand.

How about make the scheduled time configurable via environment variables? This way dev can schedule it in the next minute or so.

mcovalt:
That seems like a great approach.

fang-yan-csra:
I pushed to add the configuration. Thanks!

Issue #1002, opened 2019-07-10
modifications to QA summary page


kdionisio:
As the product owner
I would like to more efficiently view the QA summary page
So that I can more easily evaluate the overall QA status for a particular extraction script

Currently on the QA summary page (http://factotum.epa.gov/qa/extractionscript/52/summary) for an extraction script, many document records appear which do not have a QA note. 

Definition of Done:
1. 'Product' label should be removed
2.  Only include on QA summary page those documents which have a non-empty and non-null QA note.
3.  For each data document associated with the extraction script which has a non-null and non-empty QA note, the datadocument.title should be appear (no label of 'data document' is necessary, just indicate the document title)
  - datadocument.title should be a hyperlink to the data document page
   - Beneath the data document title, label 'QA note: note text note text note text'

mwfrost:
The fact that documents are being listed without any QA notes suggests that `QANote` objects are being created when they shouldn't be.

@kdionisio Can you run `SELECT * FROM dashboard_qanotes; ` and tell me if the results are full of empty-string notes?

PR #1138, opened 2019-10-15
1002 Fix QA Summary page

mwfrost:
closes: #1002 

Kathie noticed that the QA summary page was displaying `QANote` records even when no note was made. It turns out that we are creating a `QANote` object every time an `ExtractedText` record is approved, and the page was dutifully loading them all.

I have added a filter to exclude `None` and `""` notes from the page, but we might want to go back and change how the approval works. I think I built that section of code, and I presumably had a reason to create the QANote record before testing whether or not the user had filled in the note text, but it's not clear to me now. 

This branch also includes changes to how the Exit button on the QA page works. It tries to send the user back to an index page that matches how the user arrived at the QA page.


debboutr:
@mwfrost I'm looking at this PR, seems as though approving you get a QANote w/ the `qa_notes` attr set to `None`. Did you want to fix this or were you going to wait for a new ticket?

mwfrost:
I left the QANote-creation in there with hopes of fixing it on another ticket

Issue #624, opened 2018-12-13
error on data doc page for supplemental documents


kdionisio:
There is no associated data document page (http://factotum.epa.gov/datadocument/254899/) to display when a data document belongs to group type 'SD' (i.e., Supplemental documents group type), resulting in an error.

Definition of done:
For all data documents belonging to group type 'SD', the data document page should simply display the data document.

debboutr:
Here is the error:
![image](https://user-images.githubusercontent.com/7052993/49962264-aaaac500-fec9-11e8-9163-a9a0b4b31c8f.png)

this stems from there not being an `'SD'` key in the `models` dict that we use in the `get_extracted_models` function. What model/s would the 'SD' group type be extracted into? We've added no functionality for this group type to be processed, so while we can add this type as a key in the dict, we are still going to end up with  `"No Extracted Text exists for this Data Document"` on the detail page and there is no way for the text to be extracted currently, if extracting the text is even something that we want to do with this type of document.

kdionisio:
@debboutr This type of document will not have extracted text associated with it (ever). This is a catch all data group meant for things like a memo from a contractor documenting the process they used to obtain the data, etc. So while there will be data documents stored in the data group, it won't go beyond that (no extracted text, no chemicals, etc). That is why I was thinking of having the data document page for that type of data group simply display the document itself (or just display a link to the document, whichever is easier)

PR #1139, opened 2019-10-16
624 sd data doc page

justatad44:
Disabled the link to the data document detail page for supplemental document group type 'SU' and as well as the bulk data import button. To do so, I created an 'is_supplemental_doc' boolean in the data group view/model to pass to the datagroup_detail javascript.

Test makes sure a link exists for other document type ('CO'), but not for SU.


debboutr:
looking at the new DG created in this ticket @...
`http://127.0.0.1:8000/datagroup/53/`
it seems like we won't need to add the `Extracted` column in the table or the D3 donut seeing as these types of documents will never be extracted, based on what @kdionisio said in the issue #624 

Also, if a DataDocument's URL were to be punched in manually, should we have it 404? or should we not worry about it? With the link gone a user couldn't click their way to the document detail page anyway.

mwfrost:
@debboutr Thanks, I will change the test that way

Issue #1107, opened 2019-09-23
reverse match error


kdionisio:
Getting 'no reverse match' error on this data document page: http://factotum.epa.gov/datadocument/1372170/

Appears to be same error as was resolved in #1025.

Additional note - document is linked to a dsstox_lookup record that is BLANK. This is likely the cause of the error. See record dashboard_dsstoxlookup.id=10189 where sid, true_cas, and true_chemname are all blank.

Definition of Done:
- existing dsstoxlookup record with sid = blank should be removed
- all links between record dsstoxlookup.id=10189 and dashboard_rawchem should be removed
- upon upload of curated chemicals (on http://factotum.epa.gov/chemical_curation/), if a chemical record is uploaded that does not have an 'sid', that record should not be mapped to the dsstoxlookup table (essentially, if a record in the curated chemical upload file has a blank sid field in the csv, then that record is ignored in the upload)
- institute business logic in prod_factotum such that a dsstoxlookup record can not exist if sid=blank/null

debboutr:
should there be a validation when DSSToxLookup records are created or updated for them to begin w/ the text `'DTXSID'` or that they should be of a certain length beyond that beginning text?? looking at prod it seems like there was created a record w/ a `NULL` and another chem record with an `sid` of `'sid'` and the `truecas/chemname` are `'None'`.

kdionisio:
yes you can implement a validation that the 'sid' field must begin with DTXSID

debboutr:
I'm now realizing that the ticket I'm working on could be updated with using a formset with a model form built with the `BulkMuxDict` and `CSVReader` serializer to create/update DSSTox records in the `chemical_curation_index` view for validating these records. So, unless there are objections to adding this work in with the ticket.

dmlyons2:
Tested on `prod` data and still giving the error for that data document.

PR #1141, opened 2019-10-18
#1107 remove bad dsstox, use csvserializer for upload

debboutr:
I've added validators for the `sid` field in the DSSToxLookup model so that there can be no empty spaces and so that it begins w/ `DTXSID`. The ModelChoiceField for `external_id` will also fail if the INT in the CSV isn't a `pk` of a RawChem, this hadn't been failing before. I added a filter to only show DataGroups that have 1 or more uncurated chemical in the downloads dropdown of the view worked on here
`http://127.0.0.1:8000/chemical_curation/`

I did notice along the way that `updated_at` and `created_at` get populated on a `bulk_create` but on a `bulk_update` the updated_at field doesn't automagically get timestamped at the time of the operation unless done manually which I had to do here.

Other than that I remove the bad records on prod in the migration and the rest should be about the same as it was.


dashboard/views/chemical_curation.py
```diff
@@ -1,88 +1,35 @@
...
+        }
+        formset = ChemicalCurationFormSet(form_data, request.FILES)
+        if formset.is_valid():
+            ups = formset.save()
+            messages.success(request, f"{ups} records have been successfully uploaded.")
+        else:
+            errors = gather_errors(formset)
+            for e in errors:
+                messages.error(request, e)
+    return render(request, template_name, {"dg_picker_form": DataGroupSelector()})
```
mwfrost: This is a great example for me to use on my own ticket.

Issue #467, opened 2018-10-17
upload data to product model


kdionisio:
As a Factotum data curator
I would like the ability to batch upload product data for products that do not yet exist
So that I can efficiently load extracted data specific to products

Definition of Done:
1. Ability on datagroup page to batch upload product data
  - product data upload csv should accept the following columns:
    - data document ID (required; this is only column that is required to be populated)
     - dashboard_product.upc
     - dashboard_product.url
     - dashboard_product.brand_name
     - dashboard_product.size
     - dashboard_product.color
     - dashboard_product.item_id
     - dashboard_product.parent_item_id
     - dashboard_product.short_description
     - dashboard_product.long_description
     - dashboard_product.thumb_image
     - dashboard_product.medium_image
     - dashboard_product.large_image
     - dashboard_product.model_number
     - dashboard_product.manufacturer
2. Upon upload, new product(s) are created associated with the data document indicated in data document ID within the upload
3. If no UPC is included in the upload, a stub UPC is created when the new product record is created
4. If a UPC is included in the upload, application must check to ensure that UPC does not already exists. If UPC already exists, record is rejected. Upon completion of the upload, user sees a message listing all data document IDs for which the upload failed due to already existing UPCs (ie the 'rejected records').
5. Download template for above product upload csv is available on data group page; downloadable template includes all above columns, plus a column for data document filename
6. Multiple products can be created for a single document. If user attempts to upload product data for a document which already has an associated product, a new product is created with the newly uploaded data (rather than trying to insert data into the existing product record). If a user uploads a product file which includes multiple rows for a single data document ID, each row should be assumed to be a different product for that document

mwfrost:
@kdionisio Are there any Data Group types for which this feature should be disabled?

mwfrost:
This new functionality only applies to Composition data groups

mwfrost:
> Multiple products can be created for a single document. If user attempts to upload product data for a document which already has an associated product, a new product is created with the newly uploaded data (rather than trying to insert data into the existing product record). *If a user uploads a product file which includes multiple rows for a single data document ID, each row should be assumed to be a different product for that document*

@kdionisio Is it necessary to handle the case where two rows have the same data document ID _and_ identical product information? Because that second product record is going to get rejected if its UPC already exists.

mwfrost:
@kdionisio The Product model uses a `title` field - it's the headline of every product page. I don't see that among the fields listed in this ticket. Was that an intentional omission?

[...]

PR #1164, opened 2019-10-28
467 Upload products 

mwfrost:
Closes: #467 

I am still testing this with production data and an EPA-provided upload. Otherwise it is ready for review.


dashboard/forms/data_group.py
```diff
@@ -89,6 +94,128 @@ def save(self):
...
+    """
+
+
+class ProductBulkCSVFormSet(DGFormSet):
+    prefix = "products"
+    serializer = CSVReader
+
+    def __init__(self, *args, **kwargs):
+        self.form = ProductCSVForm
+        super().__init__(*args, **kwargs)
```
mcovalt: We can declare `form` above as a class variable instead of an instance variable and remove this `__init__()`.

dashboard/forms/data_group.py
```diff
@@ -89,6 +94,128 @@ def save(self):
...
+        new_products = []
+        new_productdocuments = []
+        rejected_docids = []
+        reports = ""
+        for f in self.forms:
+            f.cleaned_data["created_at"] = datetime.now()
+            product_dict = clean_dict(f.cleaned_data, Product)
+            # Only create the Product if it doesn't already exist
+            if (
+                product_dict["upc"]
```
mcovalt:
```suggestion
                product_dict.get("upc")
```
There is a possibility that this will otherwise result in a `KeyError`.

dashboard/forms/data_group.py
```diff
@@ -89,6 +94,128 @@ def save(self):
...
+    manufacturer = field_for_model(Product, "manufacturer")
+
+    def __init__(self, *args, **kwargs):
+        super(ProductCSVForm, self).__init__(*args, **kwargs)
+
+    def clean(self):
+        self.cleaned_data["upc"] = self.cleaned_data.get(
+            "upc", Product.objects.next_upc()
+        )
+        super().clean()
```
mcovalt: This should be removed. It blocks handling of `upc` in the `ProductBulkCSVFormset.save()` method. I believe it will create many duplicate stubs — the maximum ID will be the same for every product as no new products have been created when this is called.

dashboard/forms/data_group.py
```diff
@@ -89,6 +94,128 @@ def save(self):
...
+                product_dict["upc"]
+                and Product.objects.filter(upc=product_dict["upc"]).count() > 0
+            ):
+                # if the UPC is already in the database, reject
+                # this product and report it. It does not invalidate
+                # the form, though
+                rejected_docids.append(f.cleaned_data["data_document_id"].pk)
+            else:
+                if not product_dict["upc"]:
+                    product_dict["upc"] = Product.objects.next_upc()
```
mcovalt: This will never be the case since we are populating it in `ProductCSVForm.clean()`. This is the preferred place for it.

There is a possibility of race conditions to think about. Duplicate stubs will likely result if two users are in the middle of uploading simultaneously.

dashboard/forms/data_group.py
```diff
@@ -89,6 +94,128 @@ def save(self):
...
+            "large_image",
+            "model_number",
+            "manufacturer",
+        ]
+        if header != header_cols:
+            raise forms.ValidationError(f"CSV column titles should be {header_cols}")
+
+    def save(self):
+        new_products = []
+        new_productdocuments = []
```
mcovalt: Unused

dashboard/views/data_group.py
```diff
@@ -7,14 +7,17 @@
...
 
-from dashboard.forms import DataGroupForm, create_detail_formset
+from dashboard.forms import DataGroupForm
 from dashboard.forms.data_group import (
     BulkAssignProdForm,
     CleanCompFormSet,
     ExtractFileFormSet,
     UploadDocsForm,
     RegisterRecordsFormSet,
+    UploadProductsForm,
```
mcovalt: Unused.

dashboard/views/data_group.py
```diff
@@ -139,6 +149,25 @@ def data_group_detail(request, pk, template_name="data_group/datagroup_detail.ht
         else:
             context["bulkassignprod_form"] = BulkAssignProdForm(dg)
 
+    if dg.include_product_upload_form():
+        if "products-submit" in request.POST:
+            form_data = {
+                "products-TOTAL_FORMS": 0,
+                "products-INITIAL_FORMS": 0,
+                "products-MAX_NUM_FORMS": "",
+            }
```
mcovalt: This is not necessary. This data is created in the template via `{{ product_formset.management_form }}`.

dashboard/tests/functional/test_product_csv_upload.py
```diff
@@ -0,0 +1,133 @@
...
+
+class UploadProductTest(TestCase):
+    fixtures = fixtures_standard
+
+    def setUp(self):
+        self.mng_data = {
+            "cleancomp-TOTAL_FORMS": "0",
+            "cleancomp-INITIAL_FORMS": "0",
+            "cleancomp-MAX_NUM_FORMS": "",
+        }
```
mcovalt: Unused.

dashboard/views/data_group.py
```diff
@@ -139,6 +149,25 @@ def data_group_detail(request, pk, template_name="data_group/datagroup_detail.ht
...
+            product_formset = ProductBulkCSVFormSet(form_data, request.FILES)
```
mcovalt:
```suggestion
            product_formset = ProductBulkCSVFormSet(request.POST, request.FILES)
```
The whole `request.POST` should be passed in. The relevant data will be found via `ProductBulkCSVFormSet.prefix`.

mwfrost: Oh, this explains why I could not get the management form data into the formset!

dashboard/views/data_group.py
mwfrost: I can't get it to work without this in place. I removed it and got the error about the management form data.

dashboard/views/data_group.py
mwfrost: I had to roll this change back before it would work again.

dashboard/views/data_group.py
mwfrost: Rick points out that if the formset isn't being rendered in the template, then the management form won't get passed into the POST request.

mcovalt: For testing you'll have to manually pass in that data (see `test_clean_comp_data_upload.py`), but the view should use the data sent in via the template.

dashboard/forms/data_group.py
mwfrost: yikes, this is correctly identified as a problem. It's left over from an early approach

dashboard/views/data_group.py
```diff
@@ -7,14 +7,16 @@
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import render, redirect, get_object_or_404
 from django.template.defaultfilters import pluralize
+from bulkformsets.serializers import CSVReader
 
-from dashboard.forms import DataGroupForm, create_detail_formset
+from dashboard.forms import DataGroupForm
```
mcovalt:
```suggestion
from dashboard.forms import DataGroupForm, create_detail_formset
```
The view `habitsandpractices()` depends on it.

mwfrost: Committed this. Do we not have a test for catching that?

mcovalt: It looks like we don't. I'm not sure what functionality this page adds that the DataDocument detail doesn't provide.

Issue #1069, opened 2019-08-26
link to documents from PUC page


kdionisio:
As a user, 
I want to easily link to all products associated with a PUC
So that I can investigate data relevant to my PUC of interest

Definition of Done:
1. From the /puc/ page (e.g. http://factotum.epa.gov/puc/162/), the number next to 'Products' under the 'Additional Statistics' section should be a hyperlink taking you to a new page OR appearing as table at bottom of /puc/ page
2. New page/new table should have title: Products assigned to Gen cat/prod fam/prod type PUC
2. New page/new table would include a list of every product (by product.title) assigned to that PUC
3. Each product title on the page should be a hyperlink taking you to the corresponding /product/ page
4. Pagination and basic sorting functionality for list of products is requested, as number of products per PUC is continually growing (perhaps implement as table such as the one on the /datagroup/ page which includes column sorting, pagination, and ability to change # of table entries shown in each view,e .g. datatables with ajax)

PR #1167, opened 2019-10-29
#1069 remove template js using data attribute

debboutr:
@zachmunzenrider check the diff on this PR, I've been waiting for a moment to get back to trying this, but we can use the `data-*` attributes of HTML elements to pass context data through to JS. I ran the tests you made with this update, this way we don't have any JS in the template. If you're cool with it hit the merge button, a few of the changes are from linting here.


zachmunzenrider:
This looks really good, Rick. Thanks! I was pondering a good way to avoid adding that js snippet. Yours is definitely an improvement,

PR #1170, opened 2019-10-30
1028 unit type null

### LINKED ISSUE #1028: unit type null when no composition data
Currently, unit_type is a required field when uploading extracted text data for data group type composition. However, this results in records where unit_type is populated but raw_min_comp, raw_max_comp, and raw_central_comp are all null.

Definition of Done:
- if all three above named fields are null, then unit_type should be null (if user attempts to upload a unit type in the .csv upload of extracted text where all three composition fields are blank then application should not accept the upload OR application should not upload the unit type)
- for any existing raw composition data where all three raw composition fields are blank, unit type should be null

*Note that this functionality was requested in #539 (bullet 2), however after #539 was pushed to prod a user was still able to upload extracted text where unit type = unknown was specified but there was no composition data (see http://factotum.epa.gov/datadocument/1385390/)

mcovalt:
Closes #1028 

This moves validation away from the forms and into the model when possible.


mcovalt:
It looks like I've introduced some database calls. I'm going to move this back to in-progress to get them removed.

PR #1174, opened 2019-11-01
1100 log search

### LINKED ISSUE #1100: What do we search for?
As an application admin
I would like data about what users are searching for 
So that I can effectively support search configuration

Information collected:
- search string
- user or null
- date/time
- application (defaults to factotum but could include other apps in the future)

I do not want this in the same database as prod_factotum, it will go into a new database called prod_search (or something similar).[ Django Multiple Databases](https://docs.djangoproject.com/en/2.2/topics/db/multi-db/)

Def of Done: 

- A new schema that is collecting search criteria mentioned above on the production factotum application.

- Search chars stored will be 255 chars or less.

mcovalt:
Closes #1100 

A new optional setting in `settings_secret.py` called `QUERY_LOG_DATABASE` is introduced. You can set this to whichever database you want the new `QueryLog` to live on. For example, `QUERY_LOG_DATABASE = default` will continue to use your current database configuration.

You'll need to define a new database and change `QUERY_LOG_DATABASE` accordingly if you wanted `QueryLog` to live on another database. For example:

```python
DATABASES = {
    "default": {
        "ENGINE": "django.db.backends.mysql",
        "NAME": "factotum",
        "USER": "root",
        "PASSWORD": "factotum",
        "HOST": "127.0.0.1",
    },
    "log": {
        "ENGINE": "django.db.backends.mysql",
        "NAME": "factotum_log",
        "USER": "root",
        "PASSWORD": "factotum",
        "HOST": "127.0.0.1",
    },
}
QUERY_LOG_DATABASE = "log"
```
If using multiple databases, you'll have to throw in an extra command when migrating:

```shell
python manage.py migrate
python manage.py migrate --database=log
```


mwfrost:
I tested this with both options - using the default factotum schema and a separate logging schema. It works both ways, and does not capture extraneous records when clicking around in the faceted search.

PR #1179, opened 2019-11-05
1070 keywords link document

debboutr:
As seed data is set up there are only single documents that hold any given tagset that is produced in any chemical view,

[http://127.0.0.1:8000/chemical/DTXSID9020584/](http://127.0.0.1:8000/chemical/DTXSID9020584/)

 more can be found on `prod` to see what the Datatable looks like with multiple pages

[http://127.0.0.1:8000/chemical/DTXSID401015779/](http://127.0.0.1:8000/chemical/DTXSID401015779/)

 The control divs at the bottom were placed in separate rows in the event that there are many pages returned, so that the page links don't cover up the "Showing 1 to 10 of 163 entries" response.


dashboard/views/chemical.py
```diff
@@ -25,5 +50,23 @@ def chemical_detail(request, sid):
             for needle in leaflet.children:
                 if not needle.value:
                     needle.value = parent_pucs[(leaf.name, leaflet.name, needle.name)]
-    context = {"chemical": chemical, "tagDict": tagDict, "pucs": pucs}
+    context = {"chemical": chemical, "keysets": keysets, "pucs": pucs}
     return render(request, "chemicals/chemical_detail.html", context)
+
+
+@login_required()
```
mcovalt:
```suggestion
```
Should this be public since the chemical detail page is now public? Otherwise, DataTables gives warning message that the JSON is invalid (the returned response is a 302 to the login page) when a non-logged-in user clicks the buttons.

debboutr: ooohyeah, good catch here, I would think making this public would be okay, unless @dmlyons2  has an opinion to the contrary?

Issue #1202, opened 2019-11-13
link to cleaning script not present if no cleaning script


kdionisio:
Currently on the /datadocument/  page, there is a broom icon in the upper right corner linking the user to the cleaning script. This icon is present regardless of whether there actually is a cleaning script linked to the extracted text

example page: http://factotum.epa.gov/datadocument/1364904/

Definition of Done:
Broom icon representing cleaning script is only present on /datadocument/ page if there is a cleaning script (i.e. dashboard_script with script_type DC) associated with the extracted text.

PR #1218, opened 2019-11-15
1184 last edited

### LINKED ISSUE #1184: view last edited at date for chemical card - CO
As a user of Factotum data
I would like to be able to view the last edited at date for the chemical card
So that I can see if recent changes have been made to the data

Definition of Done:
**Note base this off of the audit log to get dates.**
For CO data group type, on /datadocument/ page,
- in bottom right hand corner of chemical card, include text: "Last edited 'a time ago'" (e.g. 3 seconds ago, 6 months ago, etc)
    - text should be italicized, muted, and in smaller font than any font on the rest of the chemical card
    - the 'last edited at' date should be the most recent date considering all edits to fields listed below (whether they are the original date of data entry, or a date when a field was modified/edited)
   - fields to include in the last edited at consideration:
     - extractedchemical.raw_min_comp (Raw minimum composition)
     - extractedchemical.raw_max_comp (Raw maximum composition)
     - extractedchemical.raw_central_comp (Raw central composition)
     - extractedchemical.unit_type_id (Unit type)
     - extractedchemical.report_funcuse (Reported functional use)
     - extractedchemical.ingredient_rank (Ingredient rank)
     - extractedchemical.lower_wf_analysis (Clean minimum composition)
     - extractedchemical.central_wf_analysis (Clean central composition)
     - extractedchemical.upper_wf_analysis (Clean maximum composition)
     - rawchem.raw_cas (Raw CAS)
     - rawchem.raw_chem_name (Raw chemical name)
     - rawchem.rid (DTXRID)

debboutr:
closes #1184 , #1186 , #1202 

I think it will be necessary to move the `USE_TZ` setting from `settings.py` into `setting_secret.py` to run on prod, this needs to be `True` to be aware of when now is to evaluate how long it has been since a chemical has been edited. If we leave this on `True` all of the time, we will get warnings for every fixture that holds a DataTime field like below in both tests and when loading fixtures in, plus it fails a couple of tests.
`RuntimeWarning: DateTimeField Taxonomy.updated_at received a naive datetime (2019-02-26 00:00:00) while time zone support is active.`

If you want an accurate reading of how long ago the chemical was updated, you will need to change the `USE_TZ` to `True` and then depending where you're at you might have to change the `TIME_ZONE` setting.

For ticket #1202, we just needed to tweak the context to get a script associated w/ any one of the chemicals, before we had to search the `Ingredient` model, and can now just quey the same chems for an associated script


zachmunzenrider:
On a related note: Is there a reason we can't just query off the extracted_chemical and raw_chem `updated_at` fields? It's slightly less granular than the field list that Kathie provides in the ticket, but as a practical matter it might be just fine. . . .

zachmunzenrider:
...and to add a bit more: looking at the audit log table, I see indexes on id, field_name, and user_id. At the very least we should add one for object_key. I'm not familiar with MySQL indexes, but I'd think that an object_key/model_name/field_name index would be very beneficial.

debboutr:
@kdionisio if this feature is eventually going to have a link to a modal popup it might make most sense to query the AuditLog in that request and just use the `updated_at` datetime like Zach suggests above for the initial "XX time ago" for each chemical to bring up the initial datadocument detail page. Sound good?

zachmunzenrider:
FWIW, I added this code to `AuditLog` and it made a dramatic difference in speed: 
```    class Meta:
        indexes = [
            models.Index(fields=['object_key', 'field_name', 'model_name']),
        ]
```
and also added `db_index=True` to `date_created`:
```date_created = models.DateTimeField(auto_now_add=True, db_index=True)```

Not sure if these indexes are optimal, but they made a huge difference in speed. I think we'd be fine leaving the rest of the code as-is if we add these.

kdionisio:
I leave the decision of how to get the 'last edited date' to you all - I don't care what method is used, whatever will be best long term is fine!

PR #1219, opened 2019-11-18
1183: display search count for each model

fang-yan-csra:
1. Get unique counts for new search and store data in session.
2. Display counts as badges


dashboard/tests/functional/test_search.py
```diff
@@ -78,7 +78,7 @@ def test_number_returned(self):
         response = self.client.get("/search/product/" + qs)
         response_html = html.fromstring(response.content.decode("utf8"))
         total_took = response_html.xpath('normalize-space(//*[@id="total-took"])')
-        expected_total = "7 products"  # This includes "eau" synonym records
```
mwfrost: I'm not sure why I get 8 results from this, but it keeps changing from ticket to ticket.

PR #1232, opened 2019-11-22
#908 add migration to update 2 ExtractedText records

### LINKED ISSUE #908: change associated extraction script
As the Factotum product owner
I want to change the extraction script associated with two data documents
So that I can correct a mistake made on data loading

Definition of Done:
The following two data documents:
http://factotum.epa.gov/datadocument/1372609/
http://factotum.epa.gov/datadocument/1372610/

should be associated with extraction script:
dashboard_script.title = Manual (dummy)
dashboard_script.id=14
dashboard_script.script_type=EX

 rather than the extraction script they are currently associated with.

debboutr:
closes #908 
pretty straight forward


dashboard/migrations/0139_update_documents.py
```diff
@@ -0,0 +1,22 @@
...
+    
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dashboard', '0138_field_names'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            update_extraction_script, reverse_code=migrations.RunPython.noop
```
mcovalt: If the migration did not have `reverse_code=migrations.RunPython.noop`, would a reversal throw an error? Is this a way to tell Django that nothing needs to be done for the reverse? Haven't seen this before. Seems nice.

debboutr: indeed! 
`https://docs.djangoproject.com/en/2.2/ref/migration-operations/#django.db.migrations.operations.RunPython.noop`

PR #1236, opened 2019-11-26
Removed None values on datadocument chemical list

### LINKED ISSUE #1204: remove 'none' label in chemical list
As the factotum product owner
I want to remove the 'none' label in the chemical list
to make the data document page more visually appealing

example page: http://factotum.epa.gov/datadocument/271802/

Definition of Done:
In listing of chemical records in right hand side bar of /datadocument/ page, 
- if chemical name is blank, nothing should appear for chemical name in listing rather than labeling as 'none'
- if CAS is blank, nothing should appear for CAS in chemical listing (rather than labeling as 'none')

michael-barbour:
Minor template fix to not show elements on index with no name

closes #1204


debboutr:
was the intent of this ticket to lose the scrollspy element entirely when there `raw_chem_name` and `raw_cas` are NULL?? 
no_change
# >>> 
### or should there be some default text or icon to keep the element in the scrollspy?
no_scrollspy

michael-barbour:
Per the definition of done:
if CAS is blank, nothing should appear for CAS in chemical listing (rather than labeling as 'none')

That said you're correct that this is somewhat confusing when looking at the chemical card.  I'll ask Kathie about it.

michael-barbour:
Rick, Kathie and I discussed this functionality and the plan now is to name the scrollspy element according to a hierarchy.
If the Chemical name is present use that.  If the CAS is also present use it as a subtitle.
If the Chemical name is not present, use the CAS.
If neither the Chemical name nor CAS are present use "None".

These changes have been pushed.

PR #1257, opened 2019-12-09
1224 assigned attributes not displaying

### LINKED ISSUE #1224: assigned product attributes do not appear on public /product/ pages
Highlighting of assigned product attributes does not appear on public /product/ pages.

View example here (view in both logged in and not logged in status): http://factotum.epa.gov/product/288305/

When user is logged in, user can see blue highlighting indicating that 'pods' is an assigned attribute for the product. When user is not logged in, user can *not* see blue highlighting indicating that 'pods' is an assigned attribute for the product.

Definition of Done:
User can see highlighted attributes on public /product/ page

zachmunzenrider:
Closes #1224.


michael-barbour:
I thought about this more.  If we did want to lock down this endpoint in addition to Zach's changes we could just add request.user.is_authenticated to the save conditional.

Line 301 product_curation.py: function product_detail
``` python
if request.user.is_authenticated and tagform.is_valid():
        tagform.save()
```

mwfrost:
I like Michael's suggestion, and I am going to add that conditional before I Squash & Merge this branch

Issue #1247, opened 2019-12-04
layout of chemical card


kdionisio:
As the Factotum product owner
I want to make chemical card more visually appealing
So that it is easier to look at

Definition of Done:
For CO and CP group types, in the chemical card, for both the public and password protected /datadocument/ pages,
- Remove the upper gray divider line (the line underneath the raw name and raw cas); there should only exist one gray line, to separate the raw cas/raw name/true cas/true name etc section of the card (upper) from the weight fraction/functional use/ingredient rank section of the card (lower)
- make the raw cas and raw chemical name smaller and not bold (they should not be a 'header', rather their appearance should more resemble the appearance/size of the 'Last edited at' text in the bottom right corner of the card, so that the raw cas/raw name information is not a focus of the card)
- 'upper' portion of the card (i.e. above the gray divider line) should consist of three invisible columns
  - first column should have raw chemical name in 1st row and true chemical name in 2nd row
 - second column should have raw CAS in 1st row and true CAS in 2nd row
  - third column should have the edit/delete icons in the 1st row, and DTXSID in 2nd row
 - Note text in 1st row will be the very small, muted text, while text in 2nd row will remain the size, etc it currently is

PR #1273, opened 2019-12-13
1187 view cp audit trail

justatad44:
Completed modal CP audit view with test


templates/data_document/chemical_cards.html
```diff
@@ -163,12 +163,12 @@ <h3 class="d-inline" id="ingredient_rank" title="Ingredient Rank">
                     {% if doc.data_group.is_composition or doc.data_group.is_chemical_presence and request.user.is_authenticated %}
                         <small class="text-muted float-right font-italic">
                             <a id="chemical-audit-log-{{ chemical.pk }}"
-                               {% if doc.data_group.is_composition %}
-                               title="audit log"
-                               href="{% url 'chemical_audit_log' chemical.pk %}"
-                               data-toggle="modal"
-                               data-target="#chemical-audit-log-modal"
-                               {% endif %}
+                                {% if doc.data_group.is_composition or doc.data_group.is_chemical_presence %}
```
michael-barbour: This line is going to have merge conflicts with #1247.  

I've split chemical_cards into co_cp_chemical_cards as it seems there are significant changes.  I think we can just remove this "if" on the co_cp_chemical_cards template to achieve the same effect.

michael-barbour:
One note.  This won't record changes to List Presence Keywords as those aren't stored specifically on the chemical itself.  This is outside the scope of this ticket entirely but it will probably be needed at some point as an additional audit log.

Issue #1264, opened 2019-12-11
view documents linked to a chemical


kdionisio:
As a Factotum user
I want to link directly to the documents associated with a chemical of interest
So that I can easily access data of interest

On /chemical/ page, there currently exists a table of documents, however this table only includes documents associated with the CP keywords

http://factotum.epa.gov/chemical/DTXSID7021360/

Definition of Done:
- on /chemical/ page, include table (at bottom of page) which lists *all* documents associated with the DTXSID of the chemical page user is on (this may include CP, CO, HH, etc documents).
- table should include columns for:
  - dashboard_datadocument.title (column titled 'Document')
  - dashboard_grouptype.title (column titled 'Data type')
- document.title should be a hyperlink to the appropriate /datadocument/ page
- current table listing only documents associated with CP keyword sets should be removed
- implementation of new table should be consistent with formatting and appearance of table on /datagroup/ page

debboutr:
@kdionisio did you want the Keyword Sets to be removed in this ticket as well, I get that we want to add functionality to the table, but should there still be the grouped keyword sets that we had? thanks

PR #1300, opened 2019-12-27
1264 docs per SID

mwfrost:
This builds on the approved branch for ticket #1264 , and uses the same ajax endpoint.

The 1264 branch has already been merged into this one.

As the ticket specifies, I have removed the table of documents per keyword set from the page, but the `keysets` object is still in the `response.context` for a chemical.


dashboard/tests/functional/test_ajax.py
```diff
@@ -0,0 +1,72 @@
+import json
+
+from django.test import TestCase, override_settings
+from dashboard.tests.loader import fixtures_standard
+
+
+@override_settings(ALLOWED_HOSTS=["testserver"])
+class TestProductList(TestCase):
```
michael-barbour: Looks like this is a test case for products but later on you hit endpoints /d_json/ and /c_json/ (documents and chemicals respectively).

dashboard/tests/functional/test_ajax.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def test_anonymous_read(self):
+        response = self.client.get("/products/")
+        self.assertEqual(response.status_code, 200)
+
+    def test_default_params(self):
+        response = self.client.get("/p_json/")
+        self.assertEqual(response.status_code, 200)
+        data = json.loads(response.content)
+        self.assertEquals(data["recordsTotal"], 35)
```
michael-barbour: Asserting a hard-coded number seems odd to me.  It might be better to replace this with an aggregate of all expected products `Product.objects.all().aggregate(Count('pk'))`.  If this value is going to be used repeatedly, saving it to `self` as a variable would allow us to test against it with minor changes.

This would allow us to add more products to seed data without expecting all these hard-coded values to have to change.

templates/puc/puc_detail.html
```diff
@@ -114,8 +114,7 @@ <h4>Products Assigned To "{{ puc }}"</h4>
             <tbody></tbody>
         </table>
 
-    </div>
```
michael-barbour: This div looks like it was removed on accident.

PR #1305, opened 2019-12-31
Elastic 24: Use cross_field search to treat text field as one to get better score

fang-yan-csra:


mwfrost:
Boosting the field doesn't fix the problem identified in the ticket. A PUC record that has "care" in three of its fields will always score higher on a search for "Personal Care" than the one that only has "Personal Care" in its `gencat` field. 

More explanation here:
`https://github.com/HumanExposure/factotum_elastic/issues/24#issuecomment-568112123`

If we want the PUCs to be filtered by the search term but still sorted by their IDs, we should stop building the results page from elasticsearch (which scrubs out the sorting column) and just filter a simple list of all the PUC records.

PR #1308, opened 2020-01-02
1270 docs by group type

### LINKED ISSUE #1270: figure - docs by data group type
As a Factotum user
I would like to view a plot of number of documents by data group type
So that I can easily assess the scope of documents included in the database

Definition of Done:
On home page, view plot of number of documents by data group type 
- Since each document belongs to one and only one data group type, this could be represented by a pie chart, but open to a more exciting figure as well
- Figure is visible on both logged in and public version of Factotum

Ideally figure is same as implementation in #1279

michael-barbour:
closes #1270 

There are a lot of structure decisions that I'm less than positive made sense.

I put the ajax call "grouptype_stats" in the dashboard view as that's where it is used and that's where the other dashboard stats are.  The second most likely place was ajax.py but it felt odd to separate only this stat call from the rest of the dashboard page.

The ajax route is called "grouptype/stats".  It may make sense to prepend or append something (like "api/" or "json/") to suggest that HTML won't be returned by this route.


michael-barbour:
For the sake of posterity:

I've attached the current pie chart below.
We have decided to move forward with a table representation as it will better represent the numbers visually.  Sorry for forgetting to put this in the ticket!

piechart

Issue #1266, opened 2019-12-11
summarize data in bubble plot


kdionisio:
As a Factotum user
I want to have better functionality linking data within FActotum 
So that I can easily explore the data I am interested in

Definition of Done:
- On Factotum home page, add summary 'PUC' table to left of bubble plot (same implementation and functionality as PUC table that currently exists on /chemical/ page)
- remove gen_cat labels currently overlaid on the zoomed out version of the bubble plot on the home page

kdionisio:
Amended to include removal of gen_cat labels from bubble plot on /chemical/ page as easier to remove both sets of gen_cat labels (/chemical/ and home page) at once.

PR #1309, opened 2020-01-02
REDO: 1264 documents per sid

mwfrost:
This is a redo of the previous PR for ticket #1264 . When that PR was merged, the javascript for loading the datatable was broken.


michael-barbour:
https://github.com/HumanExposure/factotum/pull/1299
Looks like I removed this code for #1266 while attempting to isolate the PUC javascript.  This restored javascript looks good and is functional.  All tests passing.

It might be worth removing the code that is the same as bubble_puc_legend.js

Issue #1278, opened 2019-12-16
figure - DTXSID by data group type


kdionisio:
As a Factotum user
I would like to view a plot of number of DTXSID by data group type
So that I can easily assess the scope of chemicals by data type included in the database

Definition of Done:
On home page, view plot of number of unique DTXSID by data group type
- Since each DTXSID may be associated with >1 data group type, this could be represented by a Venn diagram, but open to a more exciting figure as well
- Figure is visible on both logged in and public version of Factotum

mwfrost:
Just for reference:
```
RawChem.objects.filter(dsstox__isnull=False).values('dsstox__sid','extracted_text__data_document__data_group__group_type__title').annotate(Count('id'))   
```

mwfrost:
There are seven Group Type categories, which means there are theoretically 2^7=128 possible combinations of Group Types to which any SID can belong. That is _possible_ to show in a Venn Diagram, but the results look like this:
http://moebio.com/research/sevensets/

Amazing, but not very useful for us. 

A combination matrix is probably superior. Here is a fancy example: https://bost.ocks.org/mike/miserables/

kdionisio:
@mwfrost question on your comment above. That Venn diagram is very cool :-). But I see how it may not be useful. Unless the reality of our data is that we have many fewer than the 128 possible combinations? That is unclear at this point. Also possible that we could/may want to limit the Venn diagram to only a smaller number of specific group types (probably the three 'main' group types, e.g. LP, FU, CO). Other group types are much smaller and more used for a specific use case, thus not a 'core' part of the database.

In the combination matrix, what would be along the x/y axes? Group type, or chemical? I'm not visualizing how this diagram would work to satisfy the request...if the axes are chemical, then the axes may be far too large to be useful (since we have >25,000 unique DTXSIDs).

PR #1324, opened 2020-01-08
1278 Venn Diagram of chemicals per group type

mwfrost:
Closes: #1278 

There is some craziness going on in this branch. 

First, I have created a new app called `etl` for loading and maintaining ad hoc queries that we don't want to calculate at runtime. Migrate that app up to 0002 to load the model object, then run this to create its table:
```
from django.apps import apps                                                                                                   
SummarySQL = apps.get_model("etl", "SummarySQL")                                                                               
e = SummarySQL.objects.filter(table_name="summary_sid_per_grouptype").first()  
e.refresh()  
```
The table is a very stripped-down combination of bit strings and chemical counts - it gets turned into json for the Venn diagram in an ajax endpoint:
http://127.0.0.1:8000/sid_gt_json


mcovalt:
This is very cool. Do we want to integrate this into the Celery scheduler? If so, we could do the following:

In "etl/tasks.py"
```python
from celery import shared_task
from etl.models import SummarySQL

@shared_task
def refresh_summary():
    SummarySQL.refresh_all()
```

In "factotum/celery.py":
```python
...
from etl.tasks import refresh_summary
...

@app.on_after_configure.connect
def setup_periodic_tasks(sender, **kwargs):
    # Calls refresh_summary() every hour.
    sender.add_periodic_task(360.0, refresh_summary.s(), name='add every 10')
```

mwfrost:
@mcovalt Yes, that was my plan. I have stubbed in just enough that we can call it from celery later.

michael-barbour:
In order to review this, I think an overview on where you landed on scope of the venn diagram is required.
From the ticket: 

> Also possible that we could/may want to limit the Venn diagram to only a smaller number of specific group types (probably the three 'main' group types, e.g. LP, FU, CO)

This seems to not be reflected in the code.  It appears that even with valid data this plot isn't entirely correct.  The below image represents this [json](https://github.com/HumanExposure/factotum/files/4047993/sid_gt_json.txt).  Of note, my Chemical Presense is entirely contained within the [Composition/Chemical Presense] and [Composition/HHE/Chemical Presense] sets yet [Chemical Presense] and [Chemical Presense/HHE] are both visible (although the mouse-over text does not reflect that).

image

mwfrost:
I have changed the ETL app so that it no longer creates any tables in the database. Now the results of each SummarySQL query are stored in the `results` field of that model as JSON-formatted text. The `ajax.py` view gets its data from the SummarySQL model rather than looking in the database. The changes should be totally transparent to Celery.

mwfrost:
@michael-barbour That example has me wondering if it's even possible to draw an accurate diagram with this library. The diagram would look like this: 
IMG_0580

mwfrost:
@michael-barbour I plugged those example numbers into the demo tool on the developer's website: http://www.benfrederickson.com/venn-diagrams-with-d3.js/

It fails in the same way.
Screen Shot 2020-01-14 at 7 04 23 AM

mwfrost:
Matt Covalt has removed the ETL app entirely and now the expensive query is being performed through the ORM in a `@cache_page` view

PR #1341, opened 2020-01-12
1227 google analytics

### LINKED ISSUE #1227: implement google analytics
As the Factotum Product Owner
I want to be able to view statistics summarizing user behavior, site hits, etc
So that I can inform decisions on development priorities based on issues/areas/needs of the application which are getting the most use

Definition of Done:
Implementation of Google Analytics tools for Factotum

zachmunzenrider:
Closes #1227.

This turned out to be more involved than expected, as we had to make a `.env` setting visible to the Django template, which involved creating a context processor.

I added the setting ENABLE_GOOGLE_ANALYTICS to `template.env`. This should get copied over to your `.env` file for testing--set it to True, the js code should show up in the source code. Set it to False, it should not.

I'm not at all sure how to write a test for this behavior, and I'm also not 100% sure that the new setting in `drupal-compose-override.yaml` will trigger the code to be included on production. I'm also not clear on how I might test this locally. I'm going to ask Matt Covalt to review this, as he's the most intimately familiar with the code I've been changing.


factotum/environment.py
```diff
@@ -100,6 +100,11 @@ def GUNICORN_OPTS(cls):
             for k, v in entry.split("=")
         }
 
+    @property
+    def ENABLE_GOOGLE_ANALYTICS(cls):
+        default = False
+        return cls._get("ENABLE_GOOGLE_ANALYTICS", default, prefix=False)
```
mcovalt:
```suggestion
        return cls._get("ENABLE_GOOGLE_ANALYTICS", default, prefix=False) in cls.truevals
```

This will allow the variable to always normalize to `True`/`False`. Without it, the variable could be set to `"false"`, `"no"`, or `"0"`, but still evaluate to `True` because it would be a non-empty string.

PR #1361, opened 2020-01-22
1316 revise chem

### LINKED ISSUE #1316: revise chemicals API
As the product owner
I would like to update/revise the Chemicals API
so that it better conforms with RESTful API standards

Definition of Done:
For the Chemicals API, change to the following. Bulleted items are variables to return. Please label variables in the API as they are labeled below. Also, alongside each variable is the text that should be included in the API as the definition/documentation for each variable returned.

chemicalsList (should return unique list of all DTXSID in the database)
- DTXSID  (dashboard_dsstoxlookup.sid):  The DSSTox Substance Identifier, a unique identifier associated with a chemical substance.
- Preferred name (dashboard_dsstoxlookup.true_chemname): Preferred name for the chemical substance.
- Preferred CAS (dashboard_dsstoxlookup.true_cas): Preferred CAS number for the chemical substance.

chemicalsRead (for a specific DTXSID)

Main heading documentation for the Chemicals API should be:
Service providing a list of all registered chemical substances linked to data in ChemExpoDB. All chemical data in ChemExpoDB is linked by the DTXSID, or the unique structure based identifier for the chemical substance. Service provides the DTXSID, preferred chemical name, and preferred CAS.

mcovalt:
Closes #1316 
Dependent on https://github.com/HumanExposure/factotum_ws/pull/46

This makes the `Product.uber_puc` method use less SQL calls. It also fixes some intermittent testing issues related to Selenium tests and how we built paths to `MEDIA_ROOT`.


mwfrost:
I've changed the zip_file field from 100 to 300 characters - this prevents tests from failing on my local Mac, which uses long paths for temp files. 

Note that there is now a migration.

PR #1372, opened 2020-01-30
1349: data document detail page changes for HP group type

fang-yan-csra:
For HP group type, in the datadocument detail page:
1. remove the prod_name field from extracted text area
2. display organization in header
3. allow edit the organization field


dashboard/models/data_document.py
```diff
@@ -93,6 +93,10 @@ def detail_page_editable(self):
         # this could be moved to settings
         return self.data_group.group_type.code in ["CP", "HH", "CO"]
 
+    @property
+    def detail_page_include_organization(self):
+        return self.data_group.group_type.code in ["HP"]
```
debboutr: this is really just asking what group type this datadocument is in, but the title is `detail_page_blah_blah`

PR #1375, opened 2020-02-03
Auditlog trigger bugfix

zachmunzenrider:
Unticketed. This branch addresses the problem we ran into when I new model was created, but at the time the trigger SQL was being generated and run, the actual db table did not yet exist.

At Matt Frost's suggestion, I ended up overriding the migrate command (easy to do), so that once all migrations have run, it calls a function that builds and executes the sql that generates the triggers.


zachmunzenrider:
Rick--in the migrate.py override, I call `add_trigger_sql()`, which actually does run the sql, not just return the string. Also, you're right that this function will actually run every time we run `migrate`. I don't think it's a problem to drop and re-build the triggers automatically each time we migrate, but someone else might see a problem with this! I'll throw it out to them team.

debboutr:
this needs to get tested out on the `1322` branch before we can know if it will work as expected, there are no problems with the AuditLog.trigger method until after we have added the FunctionalUse model. Should we make this PR to go into 1322?

zachmunzenrider:
I'd be fine with that!

dashboard/models/audit_log.py
```diff
@@ -43,11 +50,11 @@ def get_trigger_sql(cls):
             table_name = app_label + "_" + model
             id = "id" if model == "rawchem" else "rawchem_ptr_id"
```
debboutr: this line can be improved to use model class Meta attributes to prevent problems from adding triggers to other models by using 
```
In [16]: FunctionalUse._meta.pk.attname                                                                 
Out[16]: 'id'

In [17]: ExtractedChemical._meta.pk.attname                                                             
Out[17]: 'rawchem_ptr_id'
```
`apps.all_models[app_label]` returns an ordereddict of (model_name, model), we can probably just use both elements there and then we wouldn't even necessarily need to create the intermediate dict `auditlog_fields`, just a thought, but the first part here I would say is imperative to use for future triggers.

zachmunzenrider: What a great catch! I wasn't happy with that line of code but hadn't put the time into finding a good alternative. I'll try that out

Issue #1322, opened 2020-01-08
change relationship between raw chem and functional use


kdionisio:
As a Factotum data curator
I need to load data where a single chemical record is associated with multiple functional uses
So that I can maintain relationships found in the data document.

For the FU group type, every chemical record may be associated with >=1 functional use. Currently, the data model only allows for uploading 1 functional use per chemical record.

Definition of Done:
For the FU group type, allow for each chemical record (i.e. each dashboard_rawchem record) to be associated with >=1 dashboard_extractedfunctionaluse record.
- Data model must allow for the above
- Extracted data upload on /datagroup/ page for FU group type must allow for the above
- Extracted text templates available on /datagroup/ page for FU group type must be appropriate for new data relationship

mwfrost:
This diagram shows the existing relationships and (in red) the new relationship between RawChem and ExtractedFunctionalUse.
![erd_before.png](https://images.zenhubusercontent.com/5be34a63f724c70c7221c972/22f07b33-0661-414e-9ba0-bd0fa516338e)

PR #1393, opened 2020-02-14
1322 Functional Use Schema Change

mwfrost:
Closes: #1322 

The `report_funcuse` and `clean_funcuse` attributes were once carried by models at the `RawChem` level. Ticket #1322 requires allowing multiple functional uses to be assigned to each chemical when the chemical's parent document has the "Functional Use" group type.

We solved this by adding a new `FunctionalUse` model that joins to any `RawChem` subclass. The interface allows users to add multiple `report_funcuse` values in the editing UI, and the csv upload accepts *semicolon-delimited* lists of functional uses in the `report_funcuse` field. 

Note that the `mysqlclient` dependency has been updated to version 1.4.6. Please reinstall requirements.

This schema change broke things all over the application and the tests, and required a lot of updates to the code (55 commits!). 

The branch also includes changes to how the triggers for the audit log are added to the MySQL database. Now that step happens after all migrations are done.


dashboard/forms/data_group.py
```diff
@@ -437,17 +436,42 @@ def clean(self):
...
+                            forms.ValidationError(
+                                "The reported functional use string is too long"
+                            ),
+                        )
+                    else:
+                        uses.append(FunctionalUse(report_funcuse=use))
+
             else:
                 child = None
+            if uses and self.dg.type not in ("FU",) and len(uses) > 1:
```
mwfrost: Only `ExtractedFunctionalUse` chemicals can be linked to more than one `FunctionalUse` at the moment.

dashboard/forms/data_group.py
```diff
@@ -437,17 +436,42 @@ def clean(self):
                 datadocument = None
             # Child creates
             child_params = clean_dict(data, Child)
+            use_vals = [u.strip() for u in form["report_funcuse"].value().split(";")]
```
mwfrost: When the csv upload includes more than one functional use per chemical record, the functional uses are separated by semicolons.

dashboard/migrations/0145_functionaluse.py
```diff
@@ -0,0 +1,85 @@
+# Generated by Django 2.2.4 on 2020-01-27 06:42
```
mwfrost: This includes a data migration that moves all the `report_funcuse` attributes out of existing `RawChem` subclasses and into the new model.

dashboard/models/audit_log.py
```diff
@@ -18,6 +19,26 @@ class AuditLog(models.Model):
     class Meta:
         indexes = [models.Index(fields=["object_key", "model_name", "field_name"])]
 
+    @classmethod
```
mwfrost: Every time migrations are run, the process ends by dropping and rebuilding all the audit log triggers.

dashboard/models/extracted_functional_use.py
```diff
@@ -1,14 +1,13 @@
...
+
 from .common_info import CommonInfo
 from .raw_chem import RawChem
 
 
 class ExtractedFunctionalUse(CommonInfo, RawChem):
-    report_funcuse = models.CharField(
-        "Reported functional use", max_length=255, null=True, blank=True
-    )
+    pass
```
mwfrost: This model has no attributes of its own now. It is just a subclass that allows more than one `FunctionalUse` to be assigned to a single instance of the `RawChem` parent class.

dashboard/tests/functional/test_extracted_text_upload.py
```diff
@@ -124,6 +125,32 @@ def generate_invalid_chem_csv(self):
         )
         return in_mem_sample_csv
 
+    def generate_valid_funcuse_csv(self, funcuse_string):
```
mwfrost: The argument for this method is the `report_funcuse` string, so the same method can generate a csv row with one or more functional uses, depending on the needs of the test.

dashboard/tests/functional/test_extracted_text_upload.py
```diff
@@ -271,58 +309,80 @@ def test_functionaluse_upload(self):
...
-        sample_csv_bytes = sample_csv.encode(encoding="UTF-8", errors="strict")
-        in_mem_sample_csv = InMemoryUploadedFile(
-            io.BytesIO(sample_csv_bytes),
-            field_name="extfile-bulkformsetfileupload",
-            name="Functional_use_extract_template.csv",
-            content_type="text/csv",
-            size=len(sample_csv),
-            charset="utf-8",
-        )
+        in_mem_sample_csv = self.generate_valid_funcuse_csv("surfactant;fragrance")
```
mwfrost: Here is the semicolon-delimited list of functional uses.

requirements.txt
```diff
@@ -24,7 +24,7 @@ gunicorn>=20.0.0,<20.1.0
 ipython>=7.5.0
 json-logging-py==0.2
 lxml>=4.2.2
-mysqlclient==1.4.2.post1
+mysqlclient==1.4.6
```
mwfrost: Note this change - please reinstall dependencies.

dashboard/models/audit_log.py
```diff
@@ -18,6 +19,26 @@ class AuditLog(models.Model):
...
+    def remove_trigger_sql(cls):
```
zachmunzenrider: I don't love the way this function drops all triggers in the database--I mean, it works great as long as no one adds any unrelated triggers to the db. But we can't guarantee that this will never happen. What was the problem with the prior code, `DROP TRIGGER IF EXISTS {table_name}_update_trigger;`, that was prepended to each `CREATE TRIGGER` statement?

dashboard/models/functional_use.py
```diff
@@ -0,0 +1,45 @@
...
+    A record of the `ExtractedFunctionalUse` subclass can be related to more than 
+    one `FunctionalUse` record. 
+
+    The attributes of this model used to be stored as attributes of the chemical models. 
+    Migration `0145` split it out into its own schema object and migrated the old fields
+    into the new table.
+
+    """
+
+    chem = models.ForeignKey(
```
zachmunzenrider: Life would be so much easier if we were more consistent with our naming conventions. Why is this "chem" and not "chemical"? (Although I guess this is actually consistent with the arbitrary abbreviation of "functional_use" to "funcuse".) My aged brain has to pause every time I run across these abbreviations to translate them back into the actual English name of the thing they reference. . . .

templates/core/bs4_funcuse_form.html
```diff
@@ -0,0 +1,112 @@
...
+  {% for field in form.visible_fields %}
+    {% if field.name == 'report_funcuse' %}
+      <div class="form-group mb-2 {% if not forloop.parentloop.first and not field.value and not field.errors %}
+                              mystical
+                              {% endif %}">
+      {% if forloop.parentloop.first %}
+      <span>
+        {{ field.label_tag }}
+        {% if doc.data_group.is_functional_use %}
+        <a class="btn btn-sm btn-outline-secondary hover-success float-right mb-2 mr-2"
```
zachmunzenrider: This functionality worked well, but only after I figured out I had to click this button to add an additional functional use. I wonder if labeling this "Add Functional Use" and placing it immediately below the last function use text box would be more user-friendly?

dashboard/models/audit_log.py
debboutr: the solution prior wouldn't remove the triggers if the fields were moved to a different table, i.e in `ExtractedFunctionalUse` the field being triggered moves to another model, but there will now be no trigger fields in that table that will make the SQL to DELETE those triggers and they get left behind w/o being dropped.

mwfrost: We have no other triggers in the db right now, so this is a hypothetical scenario, but it is one that is worth accounting for. Let's submit a technical debt ticket but call this branch good.

zachmunzenrider: Rick makes a valid point, though I think it wouldn't be a big deal to explicitly drop an orphaned trigger if this scenario arose.  I do still think it's a best practice to always run a precautionary `DROP TRIGGER IF EXISTS` before trying to create a new trigger.

Issue #1345, opened 2020-01-17
edit/add habits and practices data record


kdionisio:
As a Factotum data curator
I would like the ability to add/edit habits and practices data records within the UI
So that I can enter H&P data into the database

http://factotum.epa.gov/datadocument/321338/

The green '+' sign bar that is typically in the center of the page is missing, so the user can not enter any new extracted data records

Definition of Done:
On the /datadocument/ page for group type HP, the ability to add a new HP extracted data record manually. Fields available for user to enter data in the pop-up webform should be:
- dashboard_extractedhabitsandpractices.product_surveyed
- habits and practices data type (available as a drop down menu selecting from the pick list of HP data types created in #1347) (required)
- dashboard_extractedhabitsandpractices.notes
Logged in user can edit/add extracted habits and practices records, non logged in user can not

PR #1413, opened 2020-03-03
1342 revise habits and practices model

### LINKED ISSUE #1342: revise habits and practices data model
As the product owner
I have a need to revise the habits and practices data model
So that the data model better fits with changes in the priorities of scientific work

Currently, the dashboard_extractedhabitsandpractices data model contains three records - all can be deleted. 

Definition of Done:
- remove the following fields from dashboard_extractedhabitsandpractices
  - mass
   - mass_unit
   - frequency
    - frequency_unit
    - duration
     - duration_unit
     - prevalence
- each extractedhabitsandpractices record will be associated with one 'HP data type' (created in #1347); required
- a single data document may be associated with multiple extractedhabitsandpractices records

michael-barbour:
closes #1342 

Added a hotfix to `test_chemical_update` that I forgot to push to my LM model test.

No new tests as I am only removing a few fields.  The chemical card portion of this ticket seems to be in  #1345 so the data document portion will not have cards yet.


michael-barbour:
Good catch.  This seems to be an inconsistently passing test.  I'll let you know when it's updated.

michael-barbour:
It looks like the issue was the extracted text's product_name and date fields element were being read without being given a chance to populate.  I added a wait condition `text_to_be_present_in_element` so rather than fail, this will probably throw a timeout exception after 10 seconds but it works just as well.  I set a bash script and ran this test in isolation 23 times with no failures so fingers crossed :crossed_fingers:.  That said please run it again to verify it's working on your machine.

I also talked to Dave regarding my approach with migration 0149.  I am running a database operation that drops all fields within ExtractedHabitsAndPractices.  I was worried that it would cause this migration to operate as a sort of "timebomb". where the migration, if rerun, would remove valuable data in the future.  He addressed this fear by stating "I think of the migrations as the provenance for production data".  I won't remove my delete operation, but I wanted the record attached to this PR for posterity.

**TL;DR**: Test fixed and I worried about nothing!

zachmunzenrider:
Test passes now!

PR #1426, opened 2020-03-11
1376 upload product images

### LINKED ISSUE #1376: upload product images
As a data curator
I need to upload images associated with products
So that I can display the image on the /product/ page

Often, when a curator is batch uploading product data via a .csv upload on the /datagroup/ page, the curator also has the associated image files with the image of the product. There is currently not a way to upload these product images.

Definition of Done:
Ability for curator to batch upload images associated with products

*One idea for a workflow
- Curator batch uploads product data via .csv (this functionality already exists). In the batch product upload file, user indicates, for each product, the filename for the product image; filename for product image is stored
- For data groups with existing products, functionality exists on /datagroup/ page to batch upload image files (similar to how documents are uploaded). If user uploads image files, application will look for filenames which match the product image filenames previously stored

michael-barbour:
closes #1376 

This PR adds Pillow.  Django's image file package.
This PR also requires a migration

The product image workflow can be found on http://127.0.0.1:8000/datagroup/18/ within the seed data.

To get to the new code click the product tab on the upload card set.

There is now an image directory upload field (placeholder text "Choose image directory (optional)...") below the CSV upload field (placeholder text "Choose File...")

**Rules Regarding the upload directory**
- Image Directories are not required
- Images may be up to 1MB by default (controlled in the `.env`)
- Image Directories may be up to 5MB by default (controlled in the `.env`)
- Image Directories may have up to 500 images (controlled in the `.env`)

**Rules Regarding the upload process**

- Rows need not contain an image_name.  If the image_name is blank they will be processed as normal.
- If a row does contain an image_name, a matching image file with the same file name must be provided.  
- If a row contains an image_name that cannot be matched, the entire formset is deemed invalid (this allows the curators to only fix the singular row rather than having to rely on the error message to isolate potential problem rows).

**Validation**

I've added both front and back end validation.  The front end validation is simply there for the sake of the user and should help provide feedback through alerts before a faulty submit is sent.  The model validation `validate_product_image_size` is currently not in use and could be removed from this ticket.  However, it will be helpful if there is a model form that will require product.image uploads in the future.

I edited the `.env` file with 3 new variables and will be updating the wiki.  These values are:

- PRODUCT_IMAGE_DIRECTORY_MAX_UPLOAD - Maximum size of the directory upload in bytes.  Defaults to 5 MB (5,000,000 B)
- PRODUCT_IMAGE_DIRECTORY_MAX_FILE_COUNT - Maximum number of product images allowed to be uploaded at once.  Defaults to 500.
- PRODUCT_IMAGE_MAX_SIZE - Maximum size in bytes of a single product image.  Defaults to 1 MB (1,000,000 B) 

**Notes**

For your convenience I've uploaded my image directory and sample CSV.
[1376-csv-and-file-directory.zip](https://github.com/HumanExposure/factotum/files/4319253/1376-csv-and-file-directory.zip)

This works with datagroup 18 to create products and attach them to datadoc http://127.0.0.1:8000/datadocument/141765/

There is currently no place to see the image on the front end so adding `` to the product page for your newly created product should be a good way to see it on the front end.

The data group page's product tab does not differentiate the required CSV form and the optional


zachmunzenrider:
Running into an error on http://127.0.0.1:8000/product/1873/ 
ValueError at /product/1873/ :The 'file' attribute has no file associated with it.

Other products I've tried seem to work well

michael-barbour:
> Running into an error on http://127.0.0.1:8000/product/1873/
> ValueError at /product/1873/ :The 'file' attribute has no file associated with it.
> 
> Other products I've tried seem to work well

This seems to be an issue on dev as well.  I'm assuming the product page attempts to load related files for data documents but fails because you don't have the related files in your media directory.

zachmunzenrider:
Adding `` to the product page doesn't cause the image to display, even when it should (i.e., for product 1875, after I've performed the upload)

zachmunzenrider:
Also, running into a slightly familiar error when I try to upload the `dirtectory_too_large` directory. Instead of failing gracefully, I get a nasty error: `  File "C:\dev\factotum\bulkformsets\serializers.py", line 21, in __init__ raise ValueError("Unknown file type.")`

A little debugging tells me that in this situation the file type comes through as `<class 'django.core.files.temp.TemporaryFile'>`, which doesn't get caught by the `serializer.py` code

zachmunzenrider:
Everything looks good. Approving now

PR #1440, opened 2020-03-27
1412: functional use - chemical connection

fang-yan-csra:
Added endpoint /function/, which requires query parameter: document, chemical or catagory.

example url:
http://127.0.0.1:8001/function/?document=156051
http://127.0.0.1:8001/function/?chemical=DTXSID9022528
http://127.0.0.1:8001/function/?category=1


apps_api/api/tests.py
```diff
@@ -237,6 +239,40 @@ def test_serialize(self):
         )
 
 
+class TestFunction(TestCase):
```
michael-barbour: There is no test for your FuntionalUseSerializer in this endpoint.  We tend to do it on the read, but as this doesn't have a read it might be worth while pulling a random row and serializing it.
`FunctionalUseSerializer(models.FunctionalUse.objects.first()).data`

fang-yan-csra: Pushed to add RID in response data and test data fields in response as well. Please review again. Thanks!

michael-barbour:
These changes look really good.  Thanks!

PR #1449, opened 2020-04-08
1350 hp data card

### LINKED ISSUE #1350: display of HP data record card
As a Factotum data curator
I would like the ability to view data associated with each extracted HP data record
So that I can view the data I have input to Factotum

Because the HP data group type is not associated with chemicals, in place of the 'chemical cards' we typically have in the center of each /datadocument/ page, we will display a more generic 'data record' card. Each card should represented one record from the dashboard_extractedhabitsandpractices data model.

Definition of Done:
- data record cards in the center of the HP data document pages should represent each associated record in the extractedhabitsandpractices data model
- Card should generally mimic design of chemical cards used for other data group types, but with different data fields represented
- standard edit and delete icons should appear in upper right of each data card
- cards should be ordered alphabetically by product_surveyed
- Include on each data record card the following pieces of information (product_surveyed should remain the same, all other fields will change from what is currently implemented):
  - dashboard_extractedhabitsandpractices.product_surveyed (No change needed - this can stay as it is currently displayed on the HP data record card, with this field as title of the card and also this field appearing in the right hand column nav bar in lieu of chemical name)
  - HP data type associated with the data record
  - all HP tags associated with the data record (displayed as Kind: tag title); HP tags should be grouped by the HP tag 'kind'
  - dashboard_extractedhabitsandpractices.notes
  - PUCs linked to that extractedhabitsandpractices record (displayed where??)

mock-up:
[https://app.zenhub.com/files/99837877/a176c3f5-de7e-4e36-9e0e-d56bac73097c/download](https://app.zenhub.com/files/99837877/a176c3f5-de7e-4e36-9e0e-d56bac73097c/download)

michael-barbour:
Closes #1350 

Example doc at http://127.0.0.1:8000/datadocument/53/

This ticket lands in a slightly awkward place.  None of the button are working yet as those are in #1345.  This really only shows that the card is created.

New requirement "factory-boy" added and tested by building production pods.

Fixtures changed to add tags.  In order to see them run `nuke_and_pave`


zachmunzenrider:
Getting one test failure:

```Failure
Traceback (most recent call last):
  File "C:\dev\factotum\dashboard\tests\functional\test_habits_and_practices.py", line 41, in test_hp_tags
    self.assertEqual(hp1.tags.first().updated_at.day, date.today().day)
AssertionError: 8 != 13```

michael-barbour:
> This is looking good, but I think the tag/tag kind isn't set up like Kathie wants. In the ticket, all HP tags associated with the data record (displayed as Kind: tag title); HP tags should be grouped by the HP tag 'kind'
> 
> Right now you've got the tags displaying as tag title: description. I'm guessing that the kinds should be things like age, SES, and race, and the tags themselves would be things like "adult" or "infant", with "age" as the kind
> 
> Also I'm not seeing where the last bullet is implemented:
> PUCs linked to that extractedhabitsandpractices record (displayed where??)
> 
> The factory boy code looks great! I look forward to using that in an upcoming ticket. . . .

Looks like the above request was added to the issue and not the PR.  I reworked the tag to now use Kind - Name instead of Name - Definition and I've also added a section to work as a PUC link section as well as working it into the tests.

I also merged in John Deere because I had to add a migration.

PR #1469, opened 2020-04-15
1351 associate hp with puc

### LINKED ISSUE #1351: associate HP data records with PUC
As a Factotum data curator
I want to assign a PUC to each extractedhabitsandpractices data record
So that I am easily able to categorize HP documents by what type of product data they will provide

Ability to link dashboard_extractedhabitsandpractices records and PUCs is already present in the data models, however functionality does not exist in the application to create this linkage.

http://factotum.epa.gov/datadocument/321338/

Definition of Done:
- For HP data group type, on the /datadocument/ page, from an existing data record card in the center of the page, user should have ability to assign/select 1 or more PUCs to associate with that individual data record (note a PUC is associated with each data record, the PUC is not associated with the data document as a whole)
- User should also have ability to edit/change an existing PUC to HP data record association, or to remove the HP data record to PUC association
- HP extracted data record can be associated with many PUCs

michael-barbour:
closes #1351

http://127.0.0.1:8000/datadocument/53/ is a good example of a data document that uses this functionality.  The PUCs have to be set to blank to allow the bootstrap template form to parse it as "not required".

Should be a pretty straight forward to test.


AScaduto:
Everything looks great from my point of view! @mwfrost, would you like to take a look before I merge?

Issue #1443, opened 2020-04-03
batch upload data document subtitle


kdionisio:
As a FActotum data curator
I would like to upload the data document subtitle in batch mode
So that I can more efficiently load data.

Currently, there exists a 'subtitle' field in the data document model. However, the subtitle field can only be populated manually through the UI.

Definition of Done:
Ability to load data document subtitle in batch mode when user is uploading the 'register records' file to register data documents.

PR #1473, opened 2020-04-16
Added subtitle field to doc upload/modified tests

AScaduto:
Closes issue: https://github.com/HumanExposure/factotum/issues/1443

Now able to include Subtitle as a field in data doc batch upload

example: http://127.0.0.1:8000/datasource/25/datagroup_new/

image


dashboard/tests/functional/test_datadocument_create.py
```diff
@@ -49,7 +49,7 @@ def test_dd_model_with_wrong_document_type(self):
 
     def testGoodGroupTypeInCSV(self):
```
michael-barbour, line 50: I think you did everything correct on updating this test, but currently we don't have any functionality to verify fields in a csv become fields in a datadocument (ie as long as you get a data document for every row it will seem like a success).  Could you add a set of assertions to the end pulling one of these rows as an example and verifying everything gets created correctly?

AScaduto: See latest commit

Issue #1460, opened 2020-04-15
store and load EPA Registration Number (document)


kdionisio:
As a data curator
I would like the ability to store/load/edit the EPA registration number associated with a document
So that I can easily store data that will be useful for data curation

Certain documents include an official EPA registration number (e.g. https://www3.epa.gov/pesticides/chem_search/ppls/001677-00259-20190923.pdf)

Definition of Done:
- ability to store EPA document registration number associated with a document (i.e. new field in extractedtext model, max 25 characters, one EPA reg. number per document)
- ability to upload EPA document registration number associated with a document when the bulk register records csv is uploaded
- EPA document registration number should be labeled 'EPA Reg. No.' and be visible in the 'extracted text' box on the left hand side of the /datadocument/ page (also should be available on the 'edit' pop up for the extracted text portion of the /datadocument/ page)

PR #1476, opened 2020-04-20
1460 EPA registration number

mwfrost:
Closes: #1460 

This branch adds the `doc_epa_registration` attribute to the `DataDocument` model. The field is referred to as `epa_reg_number` in the csv headers because the user doesn't care that it's an attribute of the document, and it better matches the verbose name that Kathie asked for. There will be a similar attribute on the product model soon. 

There is a new migration - if you take this to review, ping me and I'll make sure it gets moved to the end of the migration chain.


mwfrost:
Anthony's branch overlapped a lot of this branch's footprint, so I have merged `sprint-release` in and reconciled all the merge conflicts. All tests pass.

dashboard/forms/data_group.py
```diff
@@ -718,6 +718,7 @@ def clean(self, *args, **kwargs):
             "url",
             "organization",
             "subtitle",
+            "epa_reg_number",
```
michael-barbour, line 721: The discrepancy between the attribute name on the form and the column name on the model seems to result in the field not actually ever getting populated in a csv upload.

Tested using `http://127.0.0.1:8000/datasource/23/datagroup_new/` and [DatagroupCreateCSV.zip](https://github.com/HumanExposure/factotum/files/4512408/DatagroupCreateCSV.zip)

dashboard/tests/functional/test_datadocument_create.py
```diff
@@ -226,7 +226,7 @@ def test_upload_html_as_datadoc(self):
 
     def test_csv_upload_row_verification(self):
         csv_string = (
-            "filename,title,document_type,url,organization,subtitle\n"
+            "filename,title,document_type,url,organization,subtitle,epa_reg_number\n"
             "Cal_Pesticide_Residues_1987.csv,Example Datadocument from CSV,SG,http://www.epa.gov,org,cpr \n"
         )
```
michael-barbour, line 232: This test has the verification that the `/datagroup_new/` endpoint's CSV actually results in a row being created.

This should probably fail if a row is null as well?  That would have caught the issue where `epa_reg_number` doesn't seem to result in a `doc_epa_registration` being created.

dashboard/forms/data_group.py
mwfrost: i have changed the name of the attribute on the model to `epa_reg_number`. It means the Product model will have its own attribute with the same name, but that's better than remembering two names for the same field here.

michael-barbour:
The changes allow for the creation of a data document with a epa reg number.  There was also functionality added for the download register records to include both subtitle and epa reg number.

The document record download provides a csv with the header "EPA reg. no." but as no change was requested on this endpoint and doesn't break functionality it seems like it's fine.

michael-barbour:
There is a migration conflict with number 0159.

PR #1478, opened 2020-04-21
1346 hp tag selector

### LINKED ISSUE #1346: associate HP tags with HP data records
As a Factotum product owner
I want to associate HP tags with HP extracted data records
So that I can categorize the records in the document

HP tags/keywords were created in #1344

CP example: http://factotum.epa.gov/datadocument/1372072/
HP example: http://factotum.epa.gov/datadocument/321338/

Definition of Done: 
Ability to, on the /datadocument/ page for HP group type, associate HP tags with HP data records.
- User should be able to assign and remove HP tags to/from individual data records on a document page (requesting same functionality and implementation as for assigning CP keywords to individual extracted chemical records on an CP data document page)
- in the autocomplete, autocomplete on/view as Kind: keyword/tag
- multiple tags may be assigned to each HP data record

michael-barbour:
closes #1346

This ticket adds a selector between `ExtractedHabitsAndPractices` and `ExtractedHabitsAndPracticesTag` on the data document page.

**Example Data Documents:**

- HP Data Doc: http://127.0.0.1:8000/datadocument/53
- CP Data Doc (similar implementation): http://127.0.0.1:8000/datadocument/354786

Tests are written in selenium.

**Factory-boy Changes**

A factory override was also created to help facilitate the building of factories.  This handles failures to create rows with faker due to unique constraints.  There are problems with this approach (I.E. the related factories get called multiple times, exit condition is max recursion depth, kwargs provided by user get overwritten after first reattempt) but it works for small batches currently and also provides a place we could add fixes for those problems if this is a strategy we want to proceed with.


zachmunzenrider:
test_habits_and_practice_cards_add_PUC is passing now; not sure why it wasn't before. And the other 2 failing tests seem completely unrelated to Michael's ticket--we're wondering if maybe they snuck in on another PR?

PR #1479, opened 2020-04-22
1464 filter docs by data group type 

### LINKED ISSUE #1464: filter docs by data group type
Underneath the 'documents' table on the /chemical/ page there is a drop down allowing user to select and filter documents based on their associated data group type. However this selection no longer appears to be working (selecting a single data group type in the drop down box no longer filters the table to only documents of that data group type).

http://factotum.epa.gov/chemical/DTXSID4020533/

Definition of done:
ability to filter/select documents in table based on existing drop down for data group type appropriately filters documents

AScaduto:
closes #1464 

Below is a good option to test but you can test on any chemical with an SID that is assigned to multiple data groups

http://127.0.0.1:8000/chemical/DTXSID9020584/


michael-barbour:
These changes look good.  Thanks!

PR #1480, opened 2020-04-22
1459: allow more than 1 functional uses for CO document type

fang-yan-csra:
1. Existing functionality to allow upload multiple report_funuse (separated by ";". E.g., "fu1; fu2") in csv. Didn't need to change there.
2. Added functionality to add more Reported Functional Use from UI, same as FU type does.
3. Modified CO chemical card to include all Functional Uses, same format as FU type.


michael-barbour:
These changes look good.  Tested with http://127.0.0.1:8000/datadocument/244036 for CO and http://127.0.0.1:8000/datadocument/5 for FU.  Also tested extracted text upload on a newly created doc Airgas with [Airgas_5_unextracted_documents.txt](https://github.com/HumanExposure/factotum/files/4524390/Airgas_5_unextracted_documents.txt)

PR #1483, opened 2020-04-28
1463 cp tag page

### LINKED ISSUE #1463: summary of CP tags and associated documents
As a data curator
I want to view a summary of extracted list presence tags and documents they have been assigned to
So that I can identify documents to review for data curation

Definition of Done:
- Summary page for each list presence tag exists
- Page includes:
  - extracted list presence tag
  - extracted list presence tag definition
  - extracted list presence tag 'kind'
  - list of all unique list presence tag 'sets' that contain that list presence tag
  - table of documents including list of all documents which have had at least one raw chemical associated with the list presence tag
- Table of documents includes
  - title of document, as a hyperlink which takes you to the /datadocument/ page

michael-barbour:
closes #1463

**Seeded Data endpoints:**
Abrasive:  http://127.0.0.1:8000/list_presence_tag/1/
Pesticide (unused in seed data):  http://127.0.0.1:8000/list_presence_tag/269/

**Production Data endpoints:**
**Pesticide:  http://127.0.0.1:8000/list_presence_tag/269/

**Checking Pesticide on production data is a good way to verify the tables populate quickly.  Pesticide has ~45,000 ExtractedListPresences associated with it and takes ~7.5 seconds to populate the table.  The complexity of `ExtractedListPresenceTag.get_tagsets()` is low so it shouldn't grow too colossally in time as the number of associations rise.  I've also written it so it only requires one database pull.


michael-barbour:
> This looks really good. I have a couple of small changes to request:
> 
>     1. I wasn't sure what the blue "Sets" button did when I first came to the page. I clicked it a few times and nothing happened. Eventually I noticed the "Documents" text to the right, and figured out they were tab headers.  :-) I recommend you add a little styling to these to make them look like the tag headers on the PUC detail page, e.g. http://127.0.0.1:8000/puc/185/, which would be both consistent and more usable
> 
>     2. It's not clear to me how the sets are ordered in the table, and it seems like there should be an order. Alphabetic, I assume?
> 
>     3. I think both tables should be sortable

1.  This critique makes sense.  I've added a border to bring the aesthetic more in line with the PUC detail page.  Hopefully this conveys the tabbed nature of the tables.  There are other nav-icons that can also convey this if this still looks unclear.  I would probably be happy with any of them https://getbootstrap.com/docs/4.0/components/navs/
image
2. Sets by default will remain unsorted when returned by `ExtractedListPresenceTags.get_tagsets()` for the sake of keeping the tagset function as quick as possible and will be sorted by the view, alphabetically by name, before returning.
3. Removed the sorting restriction to the Set table.  It does make more sense if the keys are going to be sorted.

PR #1484, opened 2020-04-28
1422 createdby updatedby

### LINKED ISSUE #1422: created by/updated by
As the Factotum product owner
I want to track who initiated and last updated data records
So that I can better track changes to the data

Definition of Done:
-  included 'created by' and 'last updated by' fields (i.e. user ID) in common info
- for LM group type (dashboard_extractedlmdoc model), add created at and updated at fields
- for all previously created records, leave created by/update by blank

zachmunzenrider:
Closes #1422. 

This ticket had some feature creep: the basic functionality is satisfied by adding these new fields to CommonInfo and adding a pre_save signal to populate them. However, I ran into lots of issues with historical data migrations, which I ultimately resolved by squashing migrations 0001-0158.

I also realized we we handling our RawChem and ExtractedText models inconsistently: whereas RawChem's child models implemented CommonInfo directly, ExtractedTest itself implemented CommonInfo, and not its child models. This caused issues with the new fields (not to mention that RawChem itself wasn't getting the CommonInfo fields at all), so I refactored RawChem and its child models to be consistent with ExtractedText.

It's important to test this ticket against prod data. You'll want to pay special attention to the data migration, where I'm moving the existing created_at and updated_at data up into RawChem from the child models. Please let me know if you'd like more guidance on how you might test this!


michael-barbour:
I'm having 4 tests fail on your branch before merging in sprint-release.  Let me know if any of these look familiar.

[Testlog_output.txt](https://github.com/HumanExposure/factotum/files/4553819/Testlog_output.txt)

mwfrost:
I'm trying to review those tests, too

zachmunzenrider:
You know, I was seeing those grouptype constraint errors, and they were driving me a little crazy. It sure seems like they should have been working fine, based on the seed data they were pulling from. I was curious when I pushed it up if they'd show up for others, or if it was part of the same mystery I was talking about in the retro. . .

michael-barbour:
`test_rawcategory_to_puc_form.py`, `test_dsstox_lookup.py`, and `test_rawcategory_to_puc.py` all are non-fixture tests that are failing on defaults not being there.  Looks like when the migrations were squashed a few model creates were lost.  (0049 and 0090 both create GroupTypes.  Running migrations prior to squash results in 2 GroupTypes being created while after squash none will be created).

I'm still investigating the audit log failing test.

dashboard/fixtures/07d_rawchem.yaml
```diff
@@ -8566,4 +10286,6 @@
     temp_id: 1
     temp_obj_name: ExtractedChemical
     rid: DTXRID00299
+    created_at: 2019-04-08 07:01:07.897509
+    created_by_id: 1
     dsstox: 12
```
michael-barbour, line 10291: The audit log test failure is due to RawChem 856 not having an updated_at being set.  This prevents the audit log from showing up.  This should resolve the last failing test.
```suggestion
    temp_id: 1
    temp_obj_name: ExtractedChemical
    rid: DTXRID00299
    created_at: 2019-04-08 07:01:07.897509
    updated_at: 2019-04-08 07:01:07.897509
    created_by_id: 1
    updated_by_id: 1
    dsstox: 12
```

PR #1502, opened 2020-05-12
display puc kind in /product/id page

BryanMG957:
Display the PUC Kind in the /product/id page.


michael-barbour:
These changes look good.

PR #1507, opened 2020-05-14
Hotfix elasticsecurity

michael-barbour:
This fixes the issue with the prod deployment of elasticsearch's x-pack security.

Currently the elasticsearch image won't pass the elastic bootstrap checks due to not enforcing SSL communication between nodes.  We have a single node cluster so this will not be an issue until we to expand.

The Factotum pod also returns a 401 on search requests due to not having the environment variables for elastic.  This is fixed by adding the factotum_elastic auth parameters to the docker-compose for factotum.


michael-barbour:
@fang-yan-csra Found a bug where due to insufficient privileges, the user logstash_internal was not allowed to update index templates.  I'll be adding this to the [wiki](https://github.com/HumanExposure/factotum/wiki/Elasticsearch-Security-X-Pack) but the role will need to have manage-index-template permissions.

To update the role simply run the below command.
```curl -XPUT "http://elarole/logstash_internal" --user "Dave" -H 'Content-Type: application/json' -d'{  "cluster": [ "manage_index_templates" ],  "indices": [    {      "names": [        "dashboard*"      ],      "privileges": [        "all"      ],      "allow_restricted_indices": false    }  ],  "applications": [],  "run_as": [],  "metadata": {},  "transient_metadata": {    "enabled": true  }}'```

PR #1508, opened 2020-05-14
1452 Add Basic JsonAPI structure

### LINKED ISSUE #1452: Change /function/ to JSON API Standard
As a product owner,
I want to conform to an API standard
So that as we build our API we have a guide to assist us and build trust with our customers

This will be a relationship on the chemical endpoint and the document endpoint.

michael-barbour:
Closes #1452 

Remember to set `ROOT_URLCONF=api`

Modified endpoint:  http://127.0.0.1:8001/functions/

This endpoint had filters for `filter[chemical]=<sid>`, `filter[document]=<id>`, `filter[category]=<id>`.  For adding the include value add the query param `include=chemical,document,category`.  These should be outlined in the [documentation](http://127.0.0.1:8001/#tag/function/paths/~1functions/get).

### Endpoint is now allows specific request fetches as well as unfiltered requests
This endpoint also used to be list only and used to not allow access without a filter query param.  I've disabled those restrictions temporarily.  I'm citing https://jsonapi.org/format/#fetching-resources as a reason specifically for the below quote, but I'd understand this is up for interpretation.

> Fetching Resources
> 
> A server MUST support fetching resource data for every URL provided as:
> 
>     a self link as part of the top-level links object
>     a self link as part of a resource-level links object
>     a related link as part of a relationship-level links object

### DSSToxLookup.SID is unusable as a primary key
This is ground that has been tread in [ChemReg](https://github.com/Chemical-Curation/chemcurator_django/pull/84#issue-396357409)
>The ID for compounds is the primary key, not the CID, due to the limitations of the djangorestframework-api package.

Without a large amount of work which would need to be maintained every DRF-JA update it seems this will be not possible.

### Resource is now called function
We need to consider what we call each one of these resources.  "function" follows the naming rules but may get confusing with "FunctionalUseCategory -> functionaluse" which is being referred to as category on the filter, include, and response

### Joining strategy
The **rid** is actually on a 1 to 1 related table **RawChem**.  I'm trying to obscure this and discreetly joining these models on the back end, but I'm afraid this may lead to issues in the future.  Likewise **True chem** and **data document** are 2 and 3 table joins away respectively.  These seem fine with the exception of sorting.


mwfrost:
I'm not sure about the endpoint names. 

I think `chemicalpresence` is an awkward choice for the collection of `ExtractedListPresenceTag` objects. It's too close to what we would presumably call the `ExtractedListPresence` endpoint. Can we change this to `chemicalpresencetag` or `chemicalpresencekeyword`?

I am comfortable with "function" being applied to the functional use records, but not when `functionaluse` is being used as the name for the `FunctionalUseCategory` objects. And now that every extracted ingredient can have a functional use (or multiple functional uses), I'm not sure where that leaves our `ExtractedFunctionalUse` model, which is just a `RawChem` that hasn't been tricked out with any other descriptive attributes. 

Here's one suggested arrangement of models and endpoints:

| Model Name     | Resource Endpoint Name |
| ----------- | ----------- |
| `DataDocument` (and its subclasses) | `dataDocument` |
| `ExtractedListPresence`      | `chemicalPresence`       |
| `ExtractedListPresenceTag`      | `chemicalPresenceTag`       |
| `ExtractedChemical`   | `composition`        |
| `FunctionalUse` | `functionalUse` |
| `FunctionalUseCategory` | `functionalUseCategory` |
| `ExtractedFunctionalUse` | maybe `chemicalFunctionalUse` ?  |

Thinking out loud about other URLs I expect us to eventually need:

`/dataDocuments/7/chemicals` the collection of `ExtractedChemical` records on `DataDocument` with pk 7
`/dataDocuments/7/chemicals/4` The `ExtractedChemical` record with pk 4
`/dataDocuments/7/chemicals/4/functionalUses` The functional uses assigned to that ingredient.

Even if we keep the names you've used, I think they need to be camelCased, so "chemicalpresence" would become "chemicalPresence."

mwfrost:
I also re-read the original ticket and saw this:
> This will be a relationship on the chemical endpoint and the document endpoint.

Right now, the docs [here](http://127.0.0.1:8001/#tag/function/paths/~1functions~1{id}~1chemical/get) 
show the chemical resources hanging on the functional use endpoint like 
`http://127.0.0.1:8001/functions/{id}/chemical`
when it should be something more like 
`/dataDocuments/{id}/relationships/functionalUses` or `/chemicals/{id}/relationships/functionalUses`.

michael-barbour:
> I'm not sure about the endpoint names.
> 
> I think `chemicalpresence` is an awkward choice for the collection of `ExtractedListPresenceTag` objects. It's too close to what we would presumably call the `ExtractedListPresence` endpoint. Can we change this to `chemicalpresencetag` or `chemicalpresencekeyword`?
> 
> I am comfortable with "function" being applied to the functional use records, but not when `functionaluse` is being used as the name for the `FunctionalUseCategory` objects. And now that every extracted ingredient can have a functional use (or multiple functional uses), I'm not sure where that leaves our `ExtractedFunctionalUse` model, which is just a `RawChem` that hasn't been tricked out with any other descriptive attributes.
> 
> Here's one suggested arrangement of models and endpoints:
> Model Name 	Resource Endpoint Name
> `DataDocument` (and its subclasses) 	`dataDocument`
> `ExtractedListPresence` 	`chemicalPresence`
> `ExtractedListPresenceTag` 	`chemicalPresenceTag`
> `ExtractedChemical` 	`composition`
> `FunctionalUse` 	`functionalUse`
> `FunctionalUseCategory` 	`functionalUseCategory`
> `ExtractedFunctionalUse` 	maybe `chemicalFunctionalUse` ?
> 
> Thinking out loud about other URLs I expect us to eventually need:
> 
> `/dataDocuments/7/chemicals` the collection of `ExtractedChemical` records on `DataDocument` with pk 7
> `/dataDocuments/7/chemicals/4` The `ExtractedChemical` record with pk 4
> `/dataDocuments/7/chemicals/4/functionalUses` The functional uses assigned to that ingredient.
> 
> Even if we keep the names you've used, I think they need to be camelCased, so "chemicalpresence" would become "chemicalPresence."

I like all these changes.  The resource conflict between "function" and "functionalUses" was the real source of my issue too.  I'll update these resources to match and their urls to match the resource being fetched.

> Right now, the docs here
> show the chemical resources hanging on the functional use endpoint like
> http://127.0.0.1:8001/functions/{id}/chemical
> when it should be something more like
> /dataDocuments/{id}/relationships/functionalUses or /chemicals/{id}/relationships/functionalUses.

This is more related to me not knowing where this ticket ends and the next begins.  I can't tell if the reverse relationships should be considered part of this ticket or the next.

I didn't want any of my work on the document or chemical endpoints or their serializers to be considered "Finished", which is why I decided not to add them to the put them into the documentation.  I believe the chemical endpoint may be functional, but I know I ran into some minor issues on the document endpoint regarding self linking and included resources.

mwfrost:
After lots of co-writing and development of other serializers, this is good to go. Documentation is up to date.

PR #1517, opened 2020-05-22
1491 faster loading chem cards

### LINKED ISSUE #1491: faster loading for docs with many chemicals
As a factotum user
I want to be able to load pages with many chemicals more quickly 
So that the page  is more user friendly

Some documents exist in Factotum that have hundreds of chemical records associated with a single document, e.g.
http://factotum.epa.gov/datadocument/1512365/

These pages take a very long time to load and are relatively non user-friendly.

Definition of Done:
- Faster loading of /datadocument/ pages which include >30-40 chemical records
- OK to change to a more appropriate organization/display of chemical cards on these pages

michael-barbour:
closes #1491

http://127.0.0.1:8000/datadocument/1512365/

This ticket's definition of done was pretty ambiguous.  Pagination of cards would greatly speed up this page's load and reduce the server toll, but due to the potential complexity of implementing it while maintaining existing functionality, I opted to skip it for this ticket.

To add pagination the scroll spy element, the way specific chemicals can be anchor linked, the card selector, and the select/deselect all buttons will need to be reworked.

There is a "lock" period on documents with a large amount of cards (3999 cards leads to a ~2 second lock) which freezes the browser after the ajax completes while the elements are being rendered to the screen.

### Loading stats:  
Sprint-Release
- CP DataDoc with 42 ELP: 943ms
- datadocument - 1512365: 79 seconds

My Ticket
- CP DataDoc with 42 ELP: initial load 318ms with a 126ms card load ajax
- datadocument - 1512365: initial load 217ms with a 14 second card load ajax


mwfrost:
As Michael explains above, this change will not make the page work smoothly with 4,000 cards, but it at least loads most of the page and explains the delay to the user. 
There is no way to effectively display and manage thousands of chemical cards in a single display, so if this solution proves inadequate to the users, the next step is to paginate the cards.

Issue #1356, opened 2020-01-22
bulk assign raw category to PUC-data display


kdionisio:
As a Factotum data curator
I want to be able to batch assign raw categories to a PUC
So that I can more easily curate data

Currently, there are many data records in Factotum with a 'raw_category' field populated. Often this raw category field provides enough information to assign any associated products to a PUC.

Definition of Done
- Create a new 'classification_method' in the dashboard_producttopuc model for 'Bulk assignment to PUC by raw category'
- For logged in users, create a new link in the 'Data Curation' tab in the nav bar for 'Bulk assign raw category to PUC'
- Create new 'Bulk assign raw category to PUC' page, which includes on page a table listing all unique data group to dashboard_datadocument.raw_category combinations
   - Table should include columns for
       - dashboard_datagroup.name (hyperlinked to /datagroup/ page)   
       - dashboard_datadocument.raw_category
       - Number of documents in that data group with that raw_category (default sort table by this column, in descending order)
  - Table should only include entries where the number of documents in the data group with that raw_category is >50
*Optional implement with AG Grid

debboutr:
Something I forgot to mention... The `raw_category` is an attr of the DataDocument. In order to link these to PUCs there will need to be a Product created? How will that work?

kdionisio:
Good call @debboutr 
We can add an additional condition on which documents appear in the above described table. A document will only appear in the table IF at least one product has been created for the document.

michael-barbour:
Are data documents without raw categories valuable?  A fair number are blank.

![image](https://user-images.githubusercontent.com/22226925/74044800-ad079280-4999-11ea-9f23-08a6e2a12ffa.png)

michael-barbour:
> Are data documents without raw categories valuable? A fair number are blank.
> 
> ![image](https://user-images.githubusercontent.com/22226925/74044800-ad079280-4999-11ea-9f23-08a6e2a12ffa.png)

Filtering out blank raw categories per Kathie's request

michael-barbour:
@kdionisio 
In regards to your previous request to filter out documents without products, do you want those filtered documents to be removed from the document count in the third column?  Should the third column be "The number of products attached to data documents by raw category"?

kdionisio:
@michael-barbour correct, documents with a blank raw category should not be counted

The 'count' should be: Number of documents in that data group with that raw_category, where raw category = null is dropped entirely.

PR #1519, opened 2020-05-27
1418 bulk assign individual products

### LINKED ISSUE #1418: refine bulk assign raw cat to PUC
As a Factotum data curator
I want to add additional features to the bulk assign raw category to PUC functionality
So that I can more efficiently and accurately use this function to curate data

http://factotum.epa.gov/bulk_rawcategory/

Definition of Done:
- change column label 'count' to 'products' (and ensure that the number displayed is the number of associated products, not the number of associated documents)
- When user goes to batch assign PUCs, user should be able to see a list of the products that are going to be assigned to the PUC (e.g. when user selects 'connect PUCs' button for a specific data group-raw category entry in the table, perhaps they are brought to a table of the relevant products such as the one displayed on /bulk_product_puc/, so that user not only can batch assign the products to a PUC but can visually scan list of the products, deselect any products that should not be assigned to the PUC, and hyperlink to the product pages to check them if needed (similar to functionality of selecting/deselecting products when we batch assign products to PUCs based on keyword)

michael-barbour:
closes #1418 

This ticket includes fixture updates to allow for easier testing on the front end.

http://127.0.0.1:8000/bulk_rawcategory/

Added a link on product count that links to the specific products.  I opted to reuse the http://127.0.0.1:8000/bulk_product_puc/ page to avoid having to write new functionality and just filtered down the results through query parameters.

Count column now product count.  The count used to be:
> Number of documents in that data group with that raw_category (default sort table by this column, in descending order) #1356

It is now the number of products linked to documents of a specific data group with a specific raw category.  This results in some changes to the original query in addition to a small speed boost.


michael-barbour:
Points 2 and 3 of Zach's above review I think are good quality of life changes.

Point 1 I'd like some confirmation on.

> 1. I found it confusing to have the "Connect PUCs" button up top, when we're really supposed to click on the product-count hyperlink to assign the PUCs. I propose getting rid of the hyperlink, moving the "Connect PUCs" button to the bottom of the table, and having it take the user to the bulk_product_puc page as the hyperlink does now.

Currently the Bulk RawCategory page has two intended workflows.  A user can click "Connect PUCs" to **Bulk Assign** all selected items or click the one of the "Product Count" hyperlinks to visit the [Bulk Product to PUC Page](http://factotum.epa.gov/bulk_product_puc/) page populated with all results in that data group with that raw category (at this point it would be considered a **Manual Batch** assignment.)

Zach's proposed workflow is to remove the multi-select and product count hyperlink.  The user then would always have to navigate to the product page and verify which products they are selecting.  This would remove the current Connect PUC workflow and simplify the focus of this page.

## Bulk RawCategory
image

## Bulk Assign Several DataGroup/RawCategory to a PUC
image

## Bulk Product to PUC 
(clicking the product count for SC Johnson above)
image

kdionisio:
@michael-barbour  From a data quality perspective, it does make sense to require the user to shift over to the 'bulk assign products to PUCs' page and at least glance through all of the products that they are about to assign (I think this is Zack's proposal). However I'm not sure this is going to be practical for some data sets. Some data sets may  have thousands of products with a single raw category - in these cases a user is not going to glance through all product names. Thus conclusion I think is that we still need both functionality options to exist (both #1 and #2 that you detail above). This is more of a business process decision on my end. I will instruct the users that if there are a 'reasonable' number of products, they should be clicking over to glance through all products (functionality #2). But above a certain 'reasonable' number, they can use functionality #1.

PR #1523, opened 2020-05-29
1514: no Bulk Create Products for HH data group

fang-yan-csra:
1. Do not show the bulk assign product form for HH group type
2. Added sql script to clear existing data as comment in ticket


dashboard/models/data_group.py
```diff
@@ -218,7 +218,7 @@ def include_clean_comp_data_form(self):
             return False
 
     def include_bulk_assign_form(self):
-        return self.datadocument_set.filter(products=None).exists()
+        return not self.is_hh and self.datadocument_set.filter(products=None).exists()
```
michael-barbour, line 221:
```suggestion
        return self.can_have_products and self.datadocument_set.filter(products=None).exists()
```

fang-yan-csra: Updated as suggested. thanks!

michael-barbour:
Looks good!

PR #1552, opened 2020-06-23
1505 remove lm fields

### LINKED ISSUE #1505: unnecessary fields on data doc page - LM
As the product owner
I want to remove unnecessary fields from the data document page for group type LM
so that I ensure data quality

For /datadocument/ pages of data group type LM, for the 'edit/add chemical record' pop up modal, this should only show fields associated with the dashboard_rawchem model (currently there are fields shown which are from the dashboard_extractedchemical model) 

http://factotum.epa.gov/datadocument/1555997/

Definition of Done:
For /datadocument/ pages of data group type LM, on the chemical card edit modal, the following fields are the only fields which should appear (remove from UI view but not from database):
- raw chemical name
- raw CAS

zachmunzenrider:
Closes #1505 

Can be tested with http://127.0.0.1:8000/datadocument/9/ . You'll need to create the Extracted Text in order to be able to create/edit chemicals.


michael-barbour:
Looks good!

PR #1557, opened 2020-06-29
Make /pucs page public

BryanMG957:
Make the /pucs page public, no login required.


BryanMG957:
Hey @zachmunzenrider , I have added the test

PR #1558, opened 2020-06-29
1527: set up prometheus metrics endpoint

fang-yan-csra:
1. Run pip install -r requirements.txt
2. Verify http://localhost:8000/metrics endpoint works for both factotum and factotum API.


michael-barbour:
This all looks good.  I built both the "Production" env and the local runserver.  I had to add an settings variable to allow the docker image to build.

I also got it running on docker-compose to verify the data was being output correctly.  No issues there.

PR #1575, opened 2020-07-13
1566 rid attribute

### LINKED ISSUE #1566: chemicals API - RID attribute
As a user of the chemicals endpoint
I want to get a list of all RIDs in Factotum
So that I can get a list of all chemical records I need to retrieve information for

Definition of Done:
For chemicals endpoint, add dashboard_rawchem.rid as an attribute

michael-barbour:
closes #1566 

127.0.0.8001/chemicalInstances/

This is a polymorphic serialized endpoint.  All results correspond to a RawChemical or one of the RC subclasses.

This is a list only endpoint.  Each of the "types" in the result set should eventually have their own self link to use as a fetch.
I.E. it makes more sense to go to `/compositions/1` rather than `chemicalInstances/1`.  As such, **fetch** requests will return a **400** with a brief error message

Includes should work for query param `include=chemical,dataDocument,products,products.puc`.


michael-barbour:
I pushed up changes to address the include issue.

DRF:JA uses a JSONRenderer that defaults to including using underscored names.  I changed that and the validator in apps_api.core.jsonapi_fixes.  It should work now.

PR #1576, opened 2020-07-13
1535: search list presence tags

fang-yan-csra:
1. Introduced django-elasticsearch-dsl (https://django-elasticsearch-dsl.readthedocs.io/en/latest/quickstart.html#search) to index django model to elastic directly, without going through logstash.
2. To re-index, run command:
$ python manage.py search_index --rebuild
ES index dashboard_tags should be created
3. When model record updates, ES document will be synced
4. Added Tags tab in global search, which will search dashboard_tags index only

This is another PR on factotum_elastic to fix the query for dashboard index:
https://github.com/HumanExposure/factotum_elastic/pull/32


zachmunzenrider:
I agree with Matt's comment here, but otherwise the functionality and tests are all working well for me, and this looks good!

mwfrost:
The elastic query does return the `puc_id` attribute - maybe the fix is as simple as switching from sorting by descending `score` to sorting by ascending `puc_id`

PR #1607, opened 2020-08-24
1591: allow upload product via API without upc

fang-yan-csra:
1. Upc is not required any more when post product via API. A generated stub_{product.id} will be used as upc when missing.
2. Moved generate stub_id upc to Product save() method to catch all paths.
3. Generate stub_id from db auto_increment call instead of aggregate max existing id as the latter does not work when no product exists yet.
4. Optimize /composition/ and /chemicalInstance API call to prefetch weight_fraction_type.


BryanMG957:
Everything looks good on my end and tests ran with no problems. I believe @zachmunzenrider wanted to take a look at it, I will leave to him if he wants to take a second look at this ticket when he back, if not I will merge.

mwfrost:
@fang-yan-csra @BryanMG957 Do either of you guys still have the example csv files that you used for testing?

fang-yan-csra:
> @fang-yan-csra @BryanMG957 Do either of you guys still have the example csv files that you used for testing?

I used postman to test this, not csv file?

PR #1621, opened 2020-08-31
1592: create duplicate product when post product with existing upc

fang-yan-csra:
When post product with existing upc via API
1. the product is created with auto-generated uuid as upc
2. duplicateproduct is created with source_upc set to the original upc

Tested with Postman.


mwfrost (approved):
Tested with Postman and test suite, new DuplicateProduct subclass record is created as required.

Issue #1597, opened 2020-07-29
include product URL in API upload


kdionisio:
As a Factotum data curator
I want to upload the product URL when I create a product via API
So that I can include all information relative to the product

Product URL is a field related to the product, which exists in products table. CUrrently, the products API does not allow the user to upload the product URL when creating a new product via API.

Definition of Done:
Ability to upload product URL when creating a new product through the products API

PR #1622, opened 2020-09-03
Upload url of product model using product_url

BryanMG957:
Will be able to upload the url using the new attribut 'product_url'


mwfrost:
Closes: #1597

zachmunzenrider (approved):
I had some initial confusion about the use of "product_url" vs the existing "url", but it turns out this was the right way to approach it, and it's working as it should.

PR #1633, opened 2020-09-15
changed compextractionscript page to reflect QA versus extracted text

michaelhuffman-gdit:
1609. Changed the lines that populated the "percent complete" area of the page. As a side note, if there are only 2 extracts and one is completed, you will get a "QA complete" and be left with a 50% completed in the percent complete column. I do not know if this was intended but this was still the case before and after to this change.


zachmunzenrider (approved):
Michael and I reviewed this together yesterday, and it looks good!

PR #1634, opened 2020-09-16
1612: include kind code when search PUC results

fang-yan-csra:
1. Include puc kind code in the puc search results.
2. Fixed test_bulk_product_puc test cases.


mwfrost (approved):
I tested both of the examples in the browser and the code was appearing as expected.

PR #1645, opened 2020-10-01
1580: industrial/occupational PUC bubble plot

fang-yan-csra:
1. Added OC PUCs bubble plot to visualizations page and chemical page
2. Synced dashboard.puc data from prod db on 9/14
To test:
1. Reload all seed data by `python manage.py nuke_and_pave`
2. For chemical page, use http://localhost:8000/chemical/DTXSID9022528/


mwfrost (approved):
Tested with new seed data, looks great.

PR #1659, opened 2020-10-13
1637: change uncurated chemical logic to rid null or empty

fang-yan-csra:
1. Changed uncurated chemical logic to rid null or empty
2. Dropped `rid` column from downloaded csv


mwfrost (approved):
The changes work as described in the ticket. Be sure to kill and restart the server - the dropdown list of data groups was not refreshing for me until I did.

Issue #1560, opened 2020-06-30
products misrepresented on PUC pages


kdionisio:
A product can be assigned to multiple PUCs via different assignment methods (e.g. manual, batch manual, etc.). In the event that a product is assigned to multiple PUCs via different assignment methods, we have defined a hierarchy which identifies the 'uber PUC" assigned to a product. That uber PUC is the PUC which is displayed on the /product/ page.

It appears that there are instances where multiple PUCs assigned to a product are being included. Specifically, w hen a product is represented on a /puc/ page.

Secifically, the product 356235 (https://factotum.epa.gov/product/356235/) (disposable convex inserts) has two PUCs assigned using different methods (Other direct contact consumer goods, and Medical/Dental). The product appears on BOTH /puc/ pages
https://factotum.epa.gov/puc/314/
https://factotum.epa.gov/puc/317/
The product should only appear on the /puc/ page for the product's 'uber' PUC.

Definition of done:
Products should only be included in the /puc/ page for the 'uber PUC' to which they are assigned (true for all tables, plots, etc on a /puc/ page)

mwfrost:
On Tuesday, September 29, Dave Lyons and I agreed that the current method for calculating the uber PUC was dangerously poor-performing once many products were assigned to any particular PUC. To clean up this technical debt, we need to add a new model nd a database relationship. We bumped the ticket up to 13 points to reflect the additional effort.

mwfrost:
To demo, find a product with multiple PUC assignments:

`select product_id, count(*) as assignments from prod_factotum.dashboard_producttopuc GROUP BY product_id HAVING assignments > 1 ORDER BY assignments DESC;`

Inspect the assignments in the database:

`select product_id, classification_method, puc_id from dashboard_producttopuc where product_id = 557935;`

Go to the product's page to see the uberpuc:
http://127.0.0.1:8000/product/557935/

Check each PUC page for the product by filtering for "Fuchsia":
http://127.0.0.1:8000/puc/1/      `<---- products with the same title show up here, but they are different records`
http://127.0.0.1:8000/puc/2/
http://127.0.0.1:8000/puc/322/    `<---- this is the only appearance of 557935`

PR #1664, opened 2020-10-26
1617: chemical detail page products table sorting

fang-yan-csra:
Chemical page Products table PUC column is retrieved from Product model uber_puc property, not a model field. I didn't figure out the django way (without using raw query) to incorporate the subquery to select uber_puc into the query. So I choose to sort by the queryset instead.
1. Override the ordering method to sort by uber_puc property when sort on PUC and PUC Kind column
2. Also fixed the Product and Associated Document sorting by using corresponding title field.


mwfrost (approved):
This looks good. The sorting solution performed adequately with production data loaded. I checked the page cited in the original ticket. 
The work being done for ticket #1560 might overlap some of these changes, but hopefully the changes are isolated enough that the 1560 refactoring will just speed things up.

PR #1665, opened 2020-10-27
Add new column

BryanMG957:


fang-yan-csra (approved):
Kind column displays as requested.

PR #1668, opened 2020-10-27
1641 implementation

michaelhuffman-gdit:


fang-yan-csra (approved):
I pushed to skip sorting on URL and View List columns.

PR #1669, opened 2020-10-27
1652 Add chemical records/data group columns

BryanMG957:


zachmunzenrider (changes requested):
These 2 columns look good, but the ticket also asks for a Date Updated column, which still needs to be added.

BryanMG957:
> These 2 columns look good, but the ticket also asks for a Date Updated column, which still needs to be added.

Hey @zachmunzenrider  Thanks for looking at the ticket. Sakshi commented on the ticket and said she no longer needed that requirement. Fang did point out to me that the sorting in the "chemical records" column is not working. I am trying to fix that.

fang-yan-csra (approved):
I pushed to eliminate the use the order_columns as those are the same as columns. Also fixed test cases. The 2 new columns look good.

PR #1671, opened 2020-10-29
1625 chem flag instructions

mwfrost:


fang-yan-csra (approved):
I pushed to make the message to be more clear. Verified the message only displays for CP data type.

Issue #1643, opened 2020-09-28
Add QA Summary page for Chemical Presence QA


Sakshi-Handa:
As a data curator, 
I would like a QA summary page for each chemical presence QA group
So that I can view summary statistics and QA notes for the documents within a QA group.

Newly created QA summary page for chemical presence data group type will be modeled after QA summary page for composition data group type (https://factotum.epa.gov/qa/compextractionscript/343/summary)

Definition of Done:
· Applies to QA for chemical presence data group types (https://factotum.epa.gov/qa/chemicalpresence/)
· Each QA summary page for chemical presence corresponds to a single chemical presence data group
· The QA Summary Page can have statistics at the top of the page (similar to the composition QA Summary page)
	• QA Group
	• Total Documents in the QA group
	• Total Approved Documents in the QA group
	• Total QA Notes
	• Total Remaining Documents (i.e. not yet QA approved) in the QA group
· Include a table with the following column headings (similar to the composition QA summary page):
	• 'Document' - name of document and hyperlink to the /datadocument/#/ page
	• 'QA Note' - the QA Note associated with the document
	• 'Audit Log' - when changes were implemented to the document, links to audit log for changes made during QA
        * 'Data Group' - name of the data group the document is associated with
        * 'Chemical Records' -  count of the total number of extracted chemical records associated with each document.

Sakshi-Handa:
Updated definition of done to include 'data group' name and 'total chemical records'. To be similar to Composition QA Summary page.

PR #1673, opened 2020-11-02
#1643: Add QA Summary page for Chemical Presence QA

fang-yan-csra:
1. Added /qa/chemicalpresencegroup/{id}/summary/ page
2. Reused the Extraction Script summary table code for the summary table in chemical presence data group as well.


mwfrost (approved):
The new page looks good, and the test is thorough.

PR #1678, opened 2020-11-05
1674: sort uber puc performance and nulls last

fang-yan-csra:
1. Introduced django-db-views library to handle database views.
2. Db migrate to create product_uber_puc view, which contains product and it's assigned uber puc. The migration script is generated by `python manage.py makeviewmigrations`
3. Added a CustomOneToOneField class that returns None instead of throw RelatedObjectDoesNotExist exception when reverse object not found.

To test, will need to apply the db migration and install the new package.
1. python manage.py migrate
2. pip install -r requirements.txt


mwfrost (approved):
Using the MySQL view is an excellent solution to this problem, and I hope it helps us fix some of our other performance problems going forward.

PR #1680, opened 2020-11-06
1638 implementation and modified test case

michaelhuffman-gdit:


fang-yan-csra (approved):
Verified on QA Data Document page,  the chemical cards are sorted by component alphabetically, then by the ranking of the chemical.

Issue #1626, opened 2020-09-09
Bulk Delete Products associated with Data Group


Sakshi-Handa:
As a data curator
I would like to efficiently remove all products within a data group
so that I can quickly and easily remove all product entries associated with a data group.

Currently, products within a data group must be deleted individually, or the entire data group has to be deleted to remove the product records. This is a problem when we need to redo extraction scripts and reupload product data, because we would create duplicate records for all products in a data group. 

Definition of Done:
•	On the /datagroup/ page, for composition documents (i.e. CO data group type), with existing products, add a trash can icon above the products column
•	After selecting the trash can icon, the user should receive a confirmation of action, including data group name and number of products that will be deleted.  
•	Action would result in the removal of all product records associated with the documents in this data group.
•	Trash can icon should include a ‘hover’ with text ‘Delete all product records associated with this data group’

mwfrost:
This process needs to be implemented as a Celery task.

mwfrost:
To demo:
http://127.0.0.1:8000/datagroup/37/
Create the products using the Bulk Create button, then delete them.

PR #1681, opened 2020-11-09
1626 delete products

mwfrost:
Closes: #1626 

This branch adds a trash can link next to the "Products" header on the Data Group detail page for Composition group types. Clicking it sends the browser to a bulk delete endpoint and starts an asynchronous Celery task to remove the related products. 

The only test is a functional one - there is no Selenium test.  

To test, use the production data if possible to replicate the real users' scale. After deleting products, they can be immediately replaced with fresh ones by using the "Bulk Create" link.


fang-yan-csra (approved):
Verified that products associated with CO data group can be all deleted by the scheduled celery task. Production data may take a long time (20+ minutes for Walmart MSDS 1) to finished.

PR #1683, opened 2020-11-10
1626: delete task after done

fang-yan-csra:
1. After delete products task done, delete it. Otherwise, data group detail page will keep polling tasks and cause UI error in console log.
2. To test, verify that after the task is done and re-direct to data group detail page, no browser console error occurs.


zachmunzenrider (approved):
Works for me!

PR #1685, opened 2020-11-10
1579 lm study type

### LINKED ISSUE #1579: LM study type
As the literature monitoring data owner
I would like to change a label on study type
So that the selection of study type is more clear

Definition of Done:
For LM data group type, change the 'Non-targeted' study type to 'Non-targeted or Suspect Screening'

zachmunzenrider:
Closes #1579.


fang-yan-csra (approved):
Verified that Study Type as option Targeted, Non-Targeted or Suspect Screening and Other as options and can be saved successfully. For local testing, use http://localhost:8000/datadocument/9/

PR #1686, opened 2020-11-10
change API pagination count

### LINKED ISSUE #1611: number of records per page
There appears to be a mismatch in the documentation vs. functionality of the Factotum APIs.

The API documentation indicate that the max number of records that can be returned per page is 1,000. However upon testing the max number of records that can be returned per page is 100.

Definition of Done:
- Documentation on max number of records returned per page and functionality of max number of records which can be returned per page need to match for all FActotum APIs
- Max number of records to return per page should be 1,000 (but can do 500 if 1,000 is unreasonable)

zachmunzenrider:
Closes #1611. 

Just a bug, so no test. I tested it locally by

1. using the prod db,
2. uncommenting `#ROOT_URLCONF=api` from `.env` and restarting the server, and
3. going to http://127.0.0.1:8000/chemicals/ 
By default the page size is 100, so should see 100 records per page and total 282 page.
4. http://127.0.0.1:8000/chemicals/?page[size]=1000
This set the page size to 1000 and you should see 1000 records per page and total 29 pages.


fang-yan-csra:
This will change the default page size to 1000 as well, don't know if that's desired behavior. To keep the default page size 100 but allow user to set page size parameter higher, we can override the JsonApiPageNumberPagination class and set max_page_size=1000. Like:

```
from rest_framework_json_api.pagination import JsonApiPageNumberPagination


class LargeJsonApiPageNumberPagination(JsonApiPageNumberPagination):
    max_page_size = 1000
```

And in settings set the pagination class to this custom one:
`"DEFAULT_PAGINATION_CLASS": "apps_api.api.custom_pagination.LargeJsonApiPageNumberPagination",`

zachmunzenrider:
I see your point! I thought that my change was what the ticket was calling for, but you're right that this is the correct solution. I'll implement this change and push it up. Thanks!,

fang-yan-csra (approved):
Verified that max page size can be set up to 1000 by using parameter page[size]=1000

Issue #1676, opened 2020-11-02
Total QA Notes Bug


Sakshi-Handa:
There appears to be an Issue with Total QA Notes statistic on /qa/compextractionscript/#/summary page.  The QA Note statistic is meant to capture when curators make changes to the document and track the number of documents that were edited in some way. (see ticket #1639)

It looks like it is counting the QA Notes for every document, and in some cases counting more QA Notes than approved documents. From what I can tell, it is adding to the count of QA notes every time a document is approved. 

For Example, Intenze Ink has 67 documents approved. There appears to be no entries in the QA Notes column. however, the "Total QA Notes" statistic shows 67. 
https://factotum.epa.gov/qa/compextractionscript/386/summary

Also, Church & Dwight shows there are 9 documents approved, but the count of QA Notes shows 10. When you sort on the document log, it looks like only 2 documents have any text written in the QA Note column. 
https://factotum.epa.gov/qa/compextractionscript/217/summary

![QA Note Count.JPG](https://images.zenhubusercontent.com/5f318c2978fa23cbc1a57b71/583de283-bff8-4343-9ee0-3bfa26fe4f78)

Definition of Done:
I would like for the "Total QA Notes" statistic to reflect the number of written QA notes associated with documents. 
The count increases when curators write text in the QA Note, and save the note to the document.

mwfrost:
If the developer sees an opportunity to prevent the creation of all these empty QANote records, that would be worth fixing on this ticket

PR #1689, opened 2020-11-13
Bug fix. Notes that are empty are now subtracted from the total amoun…

michaelhuffman-gdit:
…t of notes before displayed


mwfrost:
Closes: #1676

mwfrost (approved):
Test and code change both look good; performance on Production data is fine.

Issue #1358, opened 2020-01-22
resolve PUC assignment conflicts


kdionisio:
As a Factotum data curator
I want the ability to resolve PUC assignment conflicts
So that I can ensure I have the highest quality data

Currently, the dashboard_producttopuc model allows for a user to assign a PUC to a product using multiple different methods (e.g. manual assignment of PUC, batch assignment of PUC, batch assignment through raw category, or scripted assignment of PUC)

Definition of Done:
- Link from data curation menu in nav bar to page for 'Resolve PUC conflicts'
- Link would bring user to a page with a table listing all products which have multiple PUCs assigned, where the PUC assignments are not consistent (e.g. a product assigned to 'shampoo' PUC via manual but assigned to 'air freshener' PUC via scripted assignment)
  -Table should include columns for:
      - Product name (hyperlink to /product/ page)
      - one of the assigned PUCs (hyperlink to /PUC/ page)
      - Method of assignment for above PUC (i.e. classification method)
      - conflicting assigned PUC (hyperlink to /PUC/ page)
      - Method of assignment for above PUC

mwfrost:
After looking at this with production data, it's apparent that the database includes a lot of duplicate Product records. The table includes the Product ID in order to distinguish between cases where one product is assigned to multiple PUCs an cases where it's just two Product records with identical Title attributes.

PR #1698, opened 2020-11-19
1358 Display products with multiple PUC assignments

mwfrost:
Closes: #1358 

The ticket this closes is called "resolve PUC assignment conflicts" but the ticket only requires a page to _display_ those conflicts. The ability to delete overlapping PUC assignments will be added later. 

The page uses a new AJAX call to populate the table. I have tested it with prod data and it loads gracefully.


fang-yan-csra (approved):
Verified the Resolve PUC conflicts page displays Products with multiple PUCs assigned.

PR #1699, opened 2020-11-23
1677: changed auditlog to handle query by datadocument and chemical better

fang-yan-csra:
1. Modified auditlog model to add extracted_text_id and rawchem_id and build index on them to serve audit log on data document and chemical, without join back to current chemical, so deleted chemical can be show as well.
2. Modified auditlog model to drop a couple unused indexed (field_name, object_key + model_name + field_name). If we have query match this pattern, we can build these indices later.
3. Modified trigger to not insert new_value/old_value null/empty string pair.
4. Cleared auditlogs data to remove new_value/old_value null/empty string pair. This takes a long time on production db (about 100 minutes on my local machine)
5. Display all auditlogs instead of Update only so user can see deleted chemicals as well.

To test, from qa page, delete some chemical. then go to qa summary page, view the audit log if the data document, the delete action should show.


zachmunzenrider (approved):
Really slick and good. This looks like your changes resolve a bunch of lingering issues from the audit log creation!

PR #1700, opened 2020-11-24
1653 comp qa group layout

### LINKED ISSUE #1653: Update Composition QA Group Page Layout
As a data curator,
I would like to clean up the Composition QA Group page (/qa/compextractionscript/#/) , 
So that I can see information more relevant to the QA process for each document within a QA group. 
 
I want to remove and add columns to clean up the table corresponding to each QA Group. 
The columns "ID" and "Revision number" are distracting, so can be removed from the table display.
The 'document date' column currently represents a date extracted from the document (e.g. when the manufacturer wrote the SDS), while it would be more useful to curators to know when the document was last updated.  
**(Updated seems to refer to changes to document details, There doesn't seem to be any date that shows when chemical records in a document were last updated?)**
It would also be helpful to have a count of records in each document before starting to QA, to prioritize workflows. 
Ex. https://factotum.epa.gov/qa/compextractionscript/341/
 

Definition of done:
•	Remove the columns and values for "ID" and "Revision Number" from the /qa/compextractionscript/#/ page
•	For the "Document Date" column, replace the column header with "Updated at" and the row values with 'dashboard_extractedtext.updated_at'
o	Set the Date and Time formatting like this: "Nov. 18, 2019."
•	To the left of the "QA Checked" column, add a new column called "Chemical Records", that displays a count of the total number of extracted chemical records associated with each document. 
o	Add sorting functionality to the new "Chemical Records" column and Updated Date Column.

zachmunzenrider:
Closes #1653  

Can test against http://127.0.0.1:8000/qa/compextractionscript/12/ with dev data, or better yet with http://127.0.0.1:8000/qa/compextractionscript/341/ with prod data.

Sakshi confirmed to me that the Date Updated column refers only to the date a related chemical was most recently updated. If no chemical has been updated, as is the case with our dev data, this column will be blank, but will be populated with the correct date when a related chemical is updated. That's what my test is checking for.


fang-yan-csra (approved):
Verified qa comp extraction script detail page updated as requested.

PR #1702, opened 2020-11-25
1655 puc table

BryanMG957:


mwfrost (approved):
AG-Grid functionality looks great. Fancy column filters work. 
I added a new test and renamed the test file & class.

PR #1703, opened 2020-11-27
removed sorting block on QA Percent column

BryanMG957:


fang-yan-csra (approved):
Sort on QA Percent column works well.

PR #1705, opened 2020-11-30
Remove button if QA is complete

BryanMG957:


fang-yan-csra (approved):
Verified QA Complete displays as expected. Also changed the style to be consistent with Extraction Script QA Complete status.

PR #1710, opened 2020-12-03
1577 pmid

### LINKED ISSUE #1577: PMID
As a product owner
I want to make consolidate the PMID field
So that it is consistently stored and able to be retrieved across all data group types for which it is applicable.

Currently, pmid field exists in both the dashboard_extractedlmdoc and the dashboard_extractedhpdoc models. The PMID stored in these two fields is the same PMID.

Also, PMID is applicable to other data group types beyond LM and HP (e.g., also applicable to LP data group type).

Definition of Done:
Consolidate PMID field as property of data document, applicable to documents in data group type LM, HP, and LP

zachmunzenrider:
Closes #1577.

Requires a migration, of course. 

In addition to refactoring the pmid field up to datadocument, Sakshi agreed, this ticket would involve moving the form field out of the child forms and up to the Edit Datadocument form, where it should only appear when the datadocument belongs to a datagroup of type LM, HP, and LP.


fang-yan-csra (approved):
Verified pmid moved to datadocument model and can be edited for LP/LM/HP group types.

PR #1712, opened 2020-12-04
1696: Add Curator Note to QA Summary

fang-yan-csra:
1. Added qa_summary_note text field to datagroup and script model
2. Added endpoint point to support edit summary note via AJAX code
3. Allow user to edit summary note from QA summary page for both Ext. Script and Chemical Presence group.

To Test:
From QA drop down, choose Composition - Ext. Script or Chemical Presence. Then click on any QA Summary link in table. Verify QA Summary Note field displays current value and can be edited and then saved via the Save Note button.


zachmunzenrider (approved):
Really nice UI interface. We should re-use this for the qa chemical forms! Everything else works well and looks clean too.

PR #1714, opened 2020-12-04
1596 implementation

michaelhuffman-gdit:


zachmunzenrider (changes requested):
The ticket calls for this text to be included on the /bulk_product_puc/ page, but it is not.

zachmunzenrider (approved):

Issue #1707, opened 2020-12-01
Display PUC Classification Hierarchy


Sakshi-Handa:
As a factotum admin 
I would like to be able to view the PUC Classification methods hierarchy table
So that I can see how classification methods are ranked at any time. 

If possible, implementation would be flexible, in case we add a new classification method or want to change the hierarchy.

Definition of Done:
	• Ability to see list of PUC Classification methods via the /admin/ panel
	• Ability to see rank of each classification method in Hierarchy
          Ability to add a description of each method

PR #1715, opened 2020-12-07
1707 puc classification rebuild

mwfrost:
Closes: #1707 

I chose to solve this without a data migration to move the existing `classification_method` codes into an integer foreign key - I kept the two-letter codes in place while turning their field into the foreign key to the new `ProductToPucClassificationMethod` model. There _is_ a data migration to populate the new model, which is just a lookup table with a `rank` attribute for defining the method hierarchy.

Using a string instead of an integer for the foreign key (and for the primary key of the related lookup table) broke the fixtures that previously loaded the `ProductToPucAssignment` objects. I could not figure out how to get them to load from yaml, so every test that uses that join has to include the new `load_producttopuc` factory method. That also imposed other complications in how the tests inside a `TestCase` are rolled back.

Note: this does not implement any tiebreaking logic for PUCs that are assigned with the same method to a single product. In those cases, the UberPUC is chosen arbitrarily.

The classification methods can be viewed and edited here:
`{host URL}/admin/dashboard/producttopucclassificationmethod/`


mwfrost:
*Potential danger:*
In this branch, migration 0177 turns the `classification_method` field of the `ProductToPuc` model from a plain character string to a foreign key related to the `ProductToPucClassificationMethod` model. Django insisted on renaming the field `classification_method_id` to disambiguate calls to the key itself and calls to the target `ProductToPucClassificationMethod` object. That means that the `product_uber_puc` view has to be altered to use the new column name.

I have not been able to reverse the view migration in 0178 yet. When I am migrated all the way up to 0178 and then try to migrate down to 0177 or lower, the process fails because it tries to change the view's DDL to use a column ID that hasn't yet been restored, because the rollback from `classification_method_id` to `classification_method` doesn't happen until 0177.

I tried putting the view migration and the foreign key ALTER in the same migration file, but it still fails because the view migration is still a single operation that has to happen *after* the field change, which means its `reverse_code` is run *before* the foreign key change is reversed. As long as the view migration gets reversed first, there is always going to be some interval when the view's DDL doesn't match the the fields it refers to. 

Maybe the only way to fix this is to have a migration that deletes the view entirely? Any suggestion are welcome.

PR #1725, opened 2020-12-14
1339: curate duplicate chems within a document

fang-yan-csra:
1. Added Data Curation -> Duplicate Chemical Records page
2. Added index on rawchem (extracted_text_id, dsstox_id, component) fields to speed up the query. 
3. I didn't figure out how to use subquery of RawChem models to join with related models in a django way, so created db view model DuplicateChemicals to utilize the index in a subquery.


mwfrost:
@fang-yan-csra 
How about a `HAVING` query?
I used this on PROD and it returned the same number of records that I see on `http://127.0.0.1:8000/duplicate_chemicals_json/`:
```
from dashboard.models import DataDocument
from django.db.models import Count

qs = DataDocument.objects.filter(data_group__group_type__code='CO').values('id','extractedtext__rawchem__component','extractedtext__rawchem__dsstox__sid').annotate(chemcount=Count('extractedtext__rawchem__dsstox__sid')).filter(chemcount__gt=1)
```
This is the SQL:
```
SELECT `dashboard_datadocument`.`id`, `dashboard_rawchem`.`component`, `dashboard_dsstoxlookup`.`sid`, COUNT(`dashboard_dsstoxlookup`.`sid`) AS `chemcount` FROM `dashboard_datadocument` INNER JOIN `dashboard_datagroup` ON (`dashboard_datadocument`.`data_group_id` = `dashboard_datagroup`.`id`) INNER JOIN `dashboard_grouptype` ON (`dashboard_datagroup`.`group_type_id` = `dashboard_grouptype`.`id`) LEFT OUTER JOIN `dashboard_extractedtext` ON (`dashboard_datadocument`.`id` = `dashboard_extractedtext`.`data_document_id`) LEFT OUTER JOIN `dashboard_rawchem` ON (`dashboard_extractedtext`.`data_document_id` = `dashboard_rawchem`.`extracted_text_id`) LEFT OUTER JOIN `dashboard_dsstoxlookup` ON (`dashboard_rawchem`.`dsstox_id` = `dashboard_dsstoxlookup`.`id`) WHERE `dashboard_grouptype`.`code` = CO GROUP BY `dashboard_datadocument`.`id`, `dashboard_rawchem`.`component`, `dashboard_dsstoxlookup`.`sid` HAVING COUNT(`dashboard_dsstoxlookup`.`sid`) > 1 ORDER BY `dashboard_datadocument`.`id` DESC
```

mwfrost:
Unfortunately the performance of that GROUP BY ... HAVING query is even worse. I will keep trying other ideas.

mwfrost (commented):
I "helpfully" changed what I thought was a typo in the name of the `dashboard.views.duplcate_chemicals` method, and it broke something. Rather than troubleshoot what happened, I've reverted that commit and I'll leave the names as-is. 

The query still takes a long time on production data, but the view definitely helps. Great solution.

PR #1731, opened 2020-12-15
Only show product labels with data

BryanMG957:


mwfrost (approved):
Simple changes to Django templates - the conditional logic looks good.

PR #1733, opened 2020-12-16
1335 bubbleplot borders

BryanMG957:


fang-yan-csra (commented):
Can we make the border color the same as the PUC group belongs to instead of all Black?

BryanMG957:
> Can we make the border color the same as the PUC group belongs to instead of all Black?

Hey @fang-yan-csra Sakshi told me she wanted it black.

BryanMG957:
Made the border 4px, 1px works for the main problem (level 3 puc that are alone outside a level 1/2 circle) but Sakshi still wanted it to be visible inside a level 2 puc @fang-yan-csra

fang-yan-csra:
I'm fine with the change as the border color matches the puc group color now.

zachmunzenrider (approved):

PR #1735, opened 2020-12-18
Fix typo in duplication chemicals view

fang-yan-csra:
The PR for 1339 is already merged. This is just fixing a typo in the view name.


mwfrost (approved):
The `duplicate_chemicals` name must have overlapped some other imported method or view. This appears to disambiguate properly.

PR #1737, opened 2020-12-21
1707 retry

mwfrost:
closes: #1707 
This is a second attempt to solve ticket 1707 - my last PR went down the wrong track and had problems loading the fixtures. 

In this branch, the tests still don't work because of a combination of factors.

The `TransactionTestCase` classes (and `LiveServerTestCase` inherits `TransactionTestCase`) truncate and restore database tables after each test run. There is no new set of migrations run. So all of the `TransactionTestCases` after the first one try to reload the fixtures into a copy of the database that has been migrated once, then truncated. All of the records that are populated via data migration are removed in that truncation process. So the objects in our new  `ProductToPucClassificationMethod` are not always present when the next test tries to restore the fixtures.

The solution to this is to use the `serialized_rollback=True` parameter, which restores the database to a version that includes the data migration rows:
https://docs.djangoproject.com/en/3.1/topics/testing/overview/#rollback-emulation

 but this is causing other errors because some keys are not being reset during the rollback, as described here:
https://stackoverflow.com/questions/29226869/django-transactiontestcase-with-rollback-emulation

I thought the issue was resolved in a Django ticket five years ago, but my problem looks like the same thing: https://code.djangoproject.com/ticket/23727

I am trying to resolve it with the workaround proposed in the StackOverflow post:
```
TEST_NON_SERIALIZED_APPS = ['django.contrib.contenttypes',
                            'django.contrib.auth']
```

Any help is much appreciated.


fang-yan-csra (approved):
Verified ClassificationMethod is in own model.

PR #1739, opened 2020-12-23
1727: exclude same pucs in puc reconciliation page

fang-yan-csra:


mwfrost (approved):
I added a new functional test to confirm that the JSON reflects the ticket's specifications. Also tested in browser. Looks good!

Issue #1723, opened 2020-12-14
Delete Chemical Records in QA


Sakshi-Handa:
The delete button on the qa/extractedtext/# page is not responding.
Curators are not able to delete chemical records with the delete button. 

Ex. https://factotum.epa.gov/qa/extractedtext/1374427/
When a document has multiple chemical records with the same name, we want to delete the duplicate records. 

Definition of Done:
• Applies to chemical records on the qa/extractedtext/# page
• When a curator clicks the delete checkbox on a chemical record (when 'edit records' is active), remove the chemical record from the data document.

mwfrost:
Error message is:
`save() prohibited to prevent data loss due to unsaved related object 'chem'.`

The local variables:
```

Variable | Value
-- | --
ChildForm | <class 'django.forms.formsets.ExtractedListPresenceFormFormSet'>
ParentForm | <class 'dashboard.forms.forms.ExtractedCPCatForm'>
context | {'detail_formset': <django.forms.formsets.ExtractedListPresenceFormFormSet object at 0x00000112561B5D48>,  'doc': <DataDocument: Survey and health assessment of cosmetic products marketed as "non-preserved"- Table 5.13>,  'document_type_form': <DocumentTypeForm bound=False, valid=Unknown, fields=(document_type)>,  'ext_form': <ExtractedCPCatForm bound=False, valid=Unknown, fields=(doc_date)>,  'extracted_text': <ExtractedCPCat: >,  'nextid': 0,  'notesform': <QANotesForm bound=False, valid=Unknown, fields=(qa_notes)>,  'referer': 'qa_extraction_script',  'script': <Script: Manual (dummy)>,  'stats': '',  'unsaved': 'false'}
created | False
detail_formset | <django.forms.formsets.ExtractedListPresenceFormFormSet object at 0x0000011256434588>
doc | <DataDocument: Survey and health assessment of cosmetic products marketed as "non-preserved"- Table 5.13>
document_type_form | <DocumentTypeForm bound=False, valid=Unknown, fields=(document_type)>
exscript | <Script: Manual (dummy)>
ext_form | <ExtractedCPCatForm bound=False, valid=Unknown, fields=(doc_date)>
extext | <ExtractedCPCat: >
[...]

PR #1740, opened 2020-12-28
1723 delete chem

mwfrost:
closes: #1723 

This problem was not caused by foreign key constraints, as I suspected, but just required a change in the order the formsets were saved.


fang-yan-csra (approved):
Verified functional use chemical record can be deleted successfully.

PR #1741, opened 2020-12-29
1734: add puc classificationMethods to API resources

fang-yan-csra:
1. Added classificationMethods to API resources.
2. Modified Product Puc relation to use product_uber_puc view to review the query count.

To test:
1. Verify http://localhost:8001/classificationMethods/ returns all methods and http://localhost:8001/classificationMethods/{id} returns a single method.
2. Verify http://localhost:8001/products/ returns puc relationship as before.


BryanMG957:
Works great

PR #1749, opened 2021-01-05
1559 func use qa new

zachmunzenrider:
I have reworked the qa page for extracted text datadocuments, to re-use the chemical cards from the datadocument detail page. 

I've been testing the following doc ids against /qa/extractedtext/ and /datadocument/

            7,  # Composition
            5,  # Functional Use
            254781,  # Chemical Presence List
            354783,  # HHE Report


mwfrost:
While we were reviewing the branch, we realized that the functional use fields now appear and can be edited in the QA view for certain data group types (CP and HH) where it was not before. It looks like both CP and HH include records where a functional use has been assigned to the chemical, so we are assuming that this is technical debt, now cleaned up.
```
In [16]: GroupType.objects.filter(datagroup__datadocument__extractedtext__rawchem__functional_uses__isnull=False).distinct()
Out[16]: <QuerySet [<GroupType: Chemical presence list>, <GroupType: Composition>, <GroupType: Functional use>, <GroupType: HHE Report>]>
```

fang-yan-csra (approved):
Looks all good. Can add/delete/modify functional use records from chemical QA page.

mwfrost:
Pages to demo:
Composition 
http://127.0.0.1:8000/qa/extractedtext/1098296/

List Presence
http://127.0.0.1:8000/qa/extractedtext/1511910/

Literature Monitoring
http://127.0.0.1:8000/datadocument/1555997/

PR #1750, opened 2021-01-05
Correct QA Status Calculations

BryanMG957:
I corrected the calculations on script.py, however the calculations in that file are in my opinion not necessary. The percent of QA documents checked is already calculated in the view in order to display it in the "Percent QA Checked" column, the template already has an if/else conditional so we could simply use that conditional to choose what text is displayed as well along with the style of the button. Chose to correct the calculations in case that file serves another purpose or if was done that way to future proof something (couldn't find any other portion of the code using it though). I would also change the return from a string to something more useful.

### DIFF
--- a/dashboard/models/script.py
+++ b/dashboard/models/script.py
@@ -18,10 +18,6 @@ class Script(CommonInfo, QASummaryNote):
         ("FU", "functional use cleaning"),
     )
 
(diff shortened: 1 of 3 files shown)


fang-yan-csra (approved):
Verified QA Complete is set only if Percent QA Checked is 100%.

Issue #1282, opened 2019-12-16
synced navigation of bubble plot/table


kdionisio:
As a user
I want to more easily navigate the PUCs in the bubble plot
So that I can easily browse data

Definition of Done:
On home page and on /chemical/ page,
- Clicking on a PUC in the PUC table should automatically zoom the bubble plot in/out appropriately. 
- Clicking on a bubble in the bubble plot should automatically expand/condense the PUC table appropriately
- Applicable to both logged in and public versions of both pages

mwfrost:
Before this ticket, the front page takes 8622.61ms to completely load with the production data installed.

mwfrost:
With seed data, loading `http://127.0.0.1:8000/dl_pucs_json/?kind=FO` takes 58 ms

Issue #1377, opened 2020-02-03
chemical-PUC summary


kdionisio:
**Editing per discussion which occurred when this ticket was pointed
As a product owner
I want to create a unique link to chemical-PUC summary data
So that I can link to this data from other parts of the application and from other applications

'Chemical' in this instance always refers to the DTXSID, and not the raw chemical record.
/chemical/ page currently exists, and displays chemical-PUC summary information

Definition of Done:
On /chemical/ page, create unique URL that directs user to a specific portion of the /chemical/ page, in this case, the information on a unique chemical-PUC combination.
- unique URL exists for each chemical-PUC combination
- unique URL directs user to the /chemical/ page, where the PUC of interest is somehow 'highlighted'
   - 'filter' option for that PUC is turned 'on', such that the documents table is displaying documents associated with that PUC, and the products table is displaying products associated with that PUC
  - bubble plot is zoomed in to that PUC

zachmunzenrider:
A question about the existing filtering behavior when you arrive at this page through the new, PUC-specific URL: should the other filter icons disappear?  It might be a recipe for confusion if they do not!

Alternately, if they remain in place as they are now, and the user clicks on one, should the page just revert to its old functionality?

PR #1751, opened 2021-01-05
1282 Bubble Plot and Legend

mwfrost:
Closes: #1282 

Numerous changes all over the code, from the database to the django models to the D3 charts. 

Make sure to migrate - this branch brings in new database views for calculating the product count and the cumulative product count for each PUC, with and without a DTXSID filter applied. There is no longer any looping over Python objects - all the summaries are calculated in SQL. 

This change applies to both the Visualizations page and the Chemical Detail Page. 

Note that the Chemical Detail page can have a PUC in the URL, as specified in ticket #1377 . The page opens zoomed in on that PUC, with the legend expanded accordingly. But then the user is free to click around on other PUCs, so the URL becomes obsolete. As long as the page is meant to be a summary of the Chemical+PUC combination, this awkwardness is inevitable.

To test with the seed data, go to 

http://localhost:8000/visualizations/

and click on the PUC legend and the bubble chart. The two elements should be in sync. 

http://localhost:8000/chemical/DTXSID9022528/puc/186/


mwfrost:
The rendering of the bubble plots is still very slow with production data. The unfortunate structure of `gen_cat -> prod_fam -> prod_type` requires a `UNION` query to stitch together all the cumulative counts in SQL. The SQL views help, but it's still too expensive for my comfort. The next step in paying down this technical debt might be to implement caching somewhere along the way.

mwfrost:
There are two new migrations in this branch. One of them changes the MySQL view that was build to select the uberpuc for each product. That uberpuc is now stored as an attribute on the `ProductToPUC` model, and the view was changed to select from that field.

mwfrost:
This branch now uses the `django-cacheops` package to cache the Chemical Detail Page data into the redis key-value store. Run `pip install -r requirements.txt` before restarting your local server.

The caching happens at the DTXSID level, so each new chemical that's loaded can take up to thirty seconds (on current prod data). Subsequent page loads, though, use the cached version.

fang-yan-csra (approved):
1. Verified the puc table and bubble plot navigation are synced.
2. The chemical page initial load is slow, but consequent load is fast as queryset is cached.

Issue #1708, opened 2020-12-01
Assign Multiple Detected Chemical Flags


Sakshi-Handa:
As a data curator,
I would like to assign the detected chemical flag to multiple chemical records at once,
So that I can quickly manually assign chemical records as 'detected' or 'not detected'.

The detected chemical flag was created in #1544. It created the chem_detected_flag field and allows for curators to add the detected label to chemical records. Currently we add labels manually to each record. I would like the ability to select multiple chemical records and assign them to detected or not detected (Yes or No). 
I would like similar functionality to what is already implemented with "List Presence Keyword" tagging, where multiple records can be selected and assigned keyword tags. 
Ex. https://factotum.epa.gov/datadocument/1496345/ 
This document has 10 chemical records, with some detected and nondetects as indicated by the LP tags. 
We would like to quickly assign these records by selecting multiple at once, and labeling them with the "yes' or "no" flags. 

This is even more important when there are large amounts of detected chemicals, as in this example with 72 records:
https://factotum.epa.gov/datadocument/1496317/

![list presence keyword tagging.JPG](https://images.zenhubusercontent.com/5f318c2978fa23cbc1a57b71/5b6a0b3b-5ce2-4966-b7a4-bc1daa987783)

Definition of Done:
	• Applies to LM and LP data groups, on the /datadocument/# page
	• Apply functionality to a new modal button, titled "Detected Chemical Flag"
	• Use existing functionality within "List Presence Keywords" 
	• Ability to select "Yes" or "No" or "Unknown" and assign selected chemical records to the chem_detected_flag field
	• Ability to save selections

michaelhuffman-gdit:
Modified requirements after talking with Sakshi

mwfrost:
The product owner has agreed to accept this ticket, with the functionality only applied to List Presence documents. The Literature Monitoring documents will get this new feature on a subsequent ticket.

PR #1754, opened 2021-01-11
1708 toggle chemicals

mwfrost:
Closes: #1708 

Select multiple chemical cards on the data document page and then update the `chem_detected_flag` attribute in a batch.


dashboard/views/data_document.py
```diff
@@ -450,6 +450,47 @@ def list_presence_tag_delete(request, doc_pk, chem_pk, tag_pk):
     return redirect(url)
 
 
+# These could all be one method, just pass the yes|no|clear argument
+# The pks need to be passed, maybe as a list
+@login_required
+def detected_flag_toggle_yes(request, doc_pk):
+    #for chem_pk in pk:
+    print("reach yes")
+    chems = request.POST.get("chems")
```
mwfrost, line 459: The `chems` object that you're getting here should be a list of IDs.

dashboard/views/data_document.py
```diff
@@ -450,6 +450,47 @@ def list_presence_tag_delete(request, doc_pk, chem_pk, tag_pk):
...
+    #print(' '.join(chems))
+    #for chem_pk in chems:
+    elp = ExtractedListPresence.objects.get(pk=chems)
```
mwfrost, line 462: Passing a list of IDs to the `.get()` function does not work. The `.get()` function expects an argument that is a unique key on the model that the function is being performed on.

dashboard/views/data_document.py
```diff
@@ -450,6 +450,47 @@ def list_presence_tag_delete(request, doc_pk, chem_pk, tag_pk):
...
+    elp.chem_detected_flag=1
```
mwfrost, line 463: This attribute update must be applied to every `ExtractedListPresence` record whose primary key is in the list `chems`. That will require either a `for` loop across the selected records (potentially very bad and slow) or an `.update()` on the QuerySet of the selected records (faster). 
See here: `https://docs.djangoproject.com/en/3.1/topics/db/queries/#updating-multiple-objects-at-once`

dashboard/views/data_document.py
```diff
@@ -450,6 +450,47 @@ def list_presence_tag_delete(request, doc_pk, chem_pk, tag_pk):
     return redirect(url)
 
 
+# These could all be one method, just pass the yes|no|clear argument
```
mwfrost, line 453: Once you figure out how to apply the changes to the full set of chemicals, please combine these methods into a single method. If you find yourself writing the same code with small changes in three different methods, you have probably violated the "Don't Repeat Yourself (DRY)" design principle. `https://docs.djangoproject.com/en/3.1/misc/design-philosophies/#don-t-repeat-yourself-dry`

dashboard/views/data_document.py
```diff
@@ -450,6 +450,59 @@ def list_presence_tag_delete(request, doc_pk, chem_pk, tag_pk):
...
+#    url = reverse("data_document", args=[doc_pk])
+#    return redirect(url)
+
+# These could all be one method, just pass the yes|no|clear argument
+# The pks need to be passed, maybe as a list
+@login_required
+def detected_flag_toggle_yes(request, doc_pk):
+    chems = request.POST.getlist("chems")
+    for chem in chems: 
+        elp = ExtractedListPresence.objects.get(pk=chem)
```
mwfrost, line 477: Per Zach's suggestion, please apply this acton to the base class attribute `RawChem.chem_detected_flag` rather than to the attribute of the inheriting class.

templates/data_document/data_document_detail.html
```diff
@@ -201,6 +201,28 @@ <h5>{{ doc.data_group }}</h5>
                     {% endif %}
                 </div>
             {% endif %}
+
+            {% if tag_form and request.user.is_authenticated %}
+            {% comment %} the tag_form attribute is a good proxy for "is this a list presence document?" {% endcomment %}
```
mwfrost, line 206: This conditional logic is what needs to change. That `tag_form` data object only appears on List Presence data documents, not Literature Monitoring ones.

zachmunzenrider (approved):
Approved as per Matt's request

PR #1762, opened 2021-01-13
1732: Add PUC Classification Method to Product Resource

fang-yan-csra:
1. Modified product_uber_puc view to include all fields from dashboard_productotpuc model.
2. Modified products API to include productUberPuc relationship, which include related puc, classificationMethod, and other attributes like classification_confidence.


mwfrost (approved):
This looks good, it follows the JSON:API standard of delivering related tables with attributes as relationships in the JSON.

PR #1763, opened 2021-01-18
1743 remove compatible group types

BryanMG957:


fang-yan-csra (approved):
Verified can add/remove Compatible Group Types from Document Type admin page.

PR #1764, opened 2021-01-18
Add QA Group Count column

BryanMG957:


fang-yan-csra (approved):
Verified QA Group Documents count displays after QA begins.

PR #1765, opened 2021-01-18
Display succes message if chemical was deleted

BryanMG957:


fang-yan-csra (approved):
Verified Chemical Deleted message shows after deleting a chemical from QA ExtractedText page.

PR #1766, opened 2021-01-18
add Classification Method column

BryanMG957:


fang-yan-csra (approved):
Verified Classification Method is displayed in products table on puc page

PR #1776, opened 2021-01-26
Remove unnecessary calculation from script.py

BryanMG957:
As I mentioned in merge request:

https://github.com/HumanExposure/factotum/pull/1750

I believe the calculations in that file(script.py) are in my opinion not necessary. It took me a while to find, but that was the cause for the repeated 380+ queries, even though the info was already available in the qa.py view. I got rid of the reference to that file in the template and just used the info already available from qa.py. That part of the code in script.py is now not been used at all (or so I believe). Left it, in case I am wrong.

### DIFF
--- a/dashboard/views/qa.py
+++ b/dashboard/views/qa.py
@@ -15,7 +15,6 @@ from django.utils import timezone
 from django.utils.http import is_safe_url
 from django.utils.timesince import timesince
 from django_datatables_view.base_datatable_view import BaseDatatableView
-from django_mysql.models import add_QuerySetMixin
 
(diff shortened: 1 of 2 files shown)


BryanMG957:
Screenshot_2021-01-26_16-52-00

BryanMG957:
Screenshot_2021-01-26_16-52-18

BryanMG957:
Screenshot_2021-01-26_17-51-38

fang-yan-csra (approved):
I modified the query to make it even faster and changed the template to display Continue QA when script.qa_begun flag is set.

PR #1789, opened 2021-02-08
Change limit and pagination bulk assignment

BryanMG957:
Removing limit makes the results take too long to load (Over 500K products, each result could be thousands). After speaking with Sakshi decide to raise it instead.


fang-yan-csra (approved):
I pushed to allow show ALL of the 200 limited records on client.

PR #1794, opened 2021-02-16
QA extraction script page reduce queries

fang-yan-csra:
The current /qa/compextractionscript/{id} page works well for Functional Use type as well. Sakshi agreed to keep all the current columns. The only change I made is to reduce queries to speed up the page loading.
Also renamed the /qa/compextractionscript/ related endpoints to /qa/extractionscript/.


zachmunzenrider (approved):
Looks good. I love the name change!

PR #1795, opened 2021-02-17
1790: Functional Use Categories Page

fang-yan-csra:
Created /functional_use_categories page and link from /get_data page. This is a simple table of the current functionalusecategory model.


zachmunzenrider (approved):
The grid look a little funny with dev data, but I can tell it will look like the list presence tags grid when it's more fully populated.

PR #1799, opened 2021-02-19
1772: corret puc products count

fang-yan-csra:
1. Corrected the Products Count column in /pucs table to count assigned as uber puc only.
2. The puc detail page already used uber logic for products count and cumulative products count
3. Disable caching the productsperpuc queryset as the performance is OK and cache caused stale data displayed after the puc is deleted from a products.


mwfrost (approved):
The product counts are consistent across the displayed summary stats and the number of records in the table. Performance is fine without caching.

PR #1800, opened 2021-02-22
Fix cacheops redis connection

fang-yan-csra:
Fix the cacheops redis connection settings.


mwfrost (approved):
Tested by changing `REDIS_HOST` to a non-`localhost` value and confirming that redis could no longer connect. Testing that is breaks serves to prove that the `REDIS_HOST` variable is being loaded into the `CACHEOPS_REDIS` variable, which we had not previously provided.

PR #1816, opened 2021-02-26
1780: functional use qa summary page

fang-yan-csra:
The existing /qa/extractionscript/{id}/summary page works for Functional User type script as well. Only fix a bug that updated_at may be null caused NoneType exception.

To Test:
Go to QA -> Functional User - Ext. Script page, following the QA Summary link in the table.


mwfrost (approved):
New auditlog code works; view displays properly. I added a test for the Functional Use summary page.

PR #1822, opened 2021-03-08
1775 puc product search

### LINKED ISSUE #1775: Ability to Search by Product Name on Attribute Assignment
As a data curator
I would like the ability to search for products by name on the Bulk Assign Products to Attributes page
So that I can assign attributes to products based on the name of the product

https://factotum.epa.gov/bulk_product_tag/
Currently implemented is the ability to search by PUC (i.e. Personal Care -dental care- toothpaste) and assign attributes to the products assigned to that PUC. Since we implemented the predicted PUC model, there are many more products assigned to each PUC, and it is less feasible to assign attributes based on PUC alone. 

We would like to add an additional ability to the page to search for all products  within a PUC that have a user-specified word in the product name. (i.e. search for the word 'aerosol' and the page returns all products with the word 'aerosol' in the product name). 
Similar to how the /bulk_product_puc/ search works, but for all products in a given PUC.

*May also be worth confirming if the table is displaying the uberPUC assignment for each product, or products assigned through any method to that PUC.  

Definition of done:
	• Applies to the /bulk_product_tag/ page
	• Ability to search all products within a given PUC by user-specified keyword
	• Ability to change the number of table entries shown (50, 100, show all)

michael-barbour:
closes #1775 

Used a server-side datatable with some minor css updates.  

Table is now loaded with ajax and uses pagination starting at 100 but allows loading all rows at once.

Search uses case-insensitive contain on product title alone per the ticket.

From the parent issue:
> May also be worth confirming if the table is displaying the uberPUC assignment for each product, or products assigned through any method to that PUC.

The data was previously not filtered by uberPUC.  It is now.


mwfrost (approved):
New tests pass, functionality works as specified on prod data. Pagination and row-count features work.

PR #1824, opened 2021-03-09
1803: component label on card

mwfrost:
I have added an `if` block around the component span and included a "Component:" label. 
I believe this change to the template does not warrant a test.

To demo the change, go to any composition or chemical presence data document and confirm that when there's a component associated with a card, the component label appears. When there is no assigned component, the label should not be there.

On prod data, this document has several unassigned chemical records:
`http://127.0.0.1:8000/datadocument/172609`

Edit the cards to include a component, and the card will move to the bottom of the sort order and the component label will appear.


fang-yan-csra (approved):
Verified component label exists.

PR #1828, opened 2021-03-09
1810 display functional use category

BryanMG957:


BryanMG957:
Simple change, I was able to test by creating a relationship via the DB. From my understanding there is still no way to create the relationship through the UI. I also misunderstood Zach's 1810 ticket and thought the complete functional use model was been deleted so I wasted time trying to figure it out.

image

fang-yan-csra (approved):
Verified harmonized functional use replaced cleaned on chemical card.

PR #1834, opened 2021-03-10
1814: download composition chemical

fang-yan-csra:
Added a download button on top of /chemical page if there are CO type documents associated.

Example page: http://localhost:8000/chemical/DTXSID9022528/


michael-barbour (approved):
These changes look good.  All tests ran.  Example CSV downloaded.  Each document, product, and extracted composition data is downloaded

[dtxsid9022528_20210312.txt](https://github.com/HumanExposure/factotum/files/6131663/dtxsid9022528_20210312.txt)

PR #1835, opened 2021-03-11
1813 Functional Use Integrity

### LINKED ISSUE #1813: Enforce referential integrity between reported functional use and Functional Use Category
No FunctionalUse can be related to a FunctionalUseCategory different from one that has been related to a FunctionalUse record with the same report_funcuse string.

michael-barbour:
closes #1813 

Added validation to FunctionalUse model regarding category.  Added a test to verify the changes take place.

NOTE:  This is only applied to the clean method.  Serializers and forms with "is_valid()" will apply these changes but direct changes to the model (e.g. `FunctionalUse(report_funcuse="a", category=2)`) will not.

There was discussion on this in person that I missed, but my interpretation of talks with Matt and the ticket's original wording are as follows.  

Given this functional use
```python
base_functional_use = FunctionalUse(
    report_funcuse='a',
    category=1
)
```

then functional uses associated with the same report_funcuse but no category would be valid
```python
valid_functional_use = FunctionalUse(
    report_funcuse='a',
    category=None
)
```
but a functional use associated to a new category would be invalid
```python
invalid_functional_use = FunctionalUse(
    report_funcuse='a',
    category=2
)
```


mwfrost (approved):
Because there is no affordance in the UI yet for assigning and editing functional use category (aka "harmonized functional use") to functional use records, I tested this new validator in the admin page. I also added the category and the chemical to the admin list view for functional uses

Issue #1807, opened 2021-02-24
Functional Use Category Assignment Pt. 1 


Sakshi-Handa:
Pt. 1 will create the table of distinct 'reported functional uses'.

As a factotum curator
I would like to assign unique functional uses to the Func. Use Categories
So that I can harmonize 'reported functional uses' to their respective categories.

Reported Functional uses are currently stored in factotum such that every functional use entry appears, including duplicates. 
I would like to create a table that reports only the unique reported functional uses, so that different spellings would be considered 1 distinct value. (i.e. multiple entries of 'FRAGRANCE', 'Fragrance', 'fragrance' would all be considered one item to be curated.)
There are over 200k functional use entries, but only 4.8k are distinct free-text values.

Query From Matt: SELECT fu.report_funcuse , count(*) FROM factotum_prod.dashboard_functionaluse fu GROUP BY report_funcuse;
![image.png](https://images.zenhubusercontent.com/5f318c2978fa23cbc1a57b71/e6e872c8-7e8e-4ca8-8fe0-f9d369cfb88c)

Definition of Done:
	• New page for the curation of 'reported functional uses' to 'Functional Use Categories'
	• Display in table 'distinct free-text values' of reported functional uses
	• Add a column to display 'Assigned functional use category', which will display the functional use category once it is assigned   
	• Ability to search the table by keyword

mwfrost:
In the pointing meeting, the product owner specified that each reported functional use string will only apply to a single functional use category, regardless of its different chemicals or sources.

michael-barbour:
Available to authenticated users as "Functional Use Curation" on the navbar drop down for "Data Curation" or at `/functional_use_curation/`

PR #1836, opened 2021-03-15
1807 Display distinct reported functional use+category assignments

mwfrost:
Closes: #1807 

Adds an ag-grid table of distinct reported functional use + functional use category values.


zachmunzenrider (approved):
Looks great, passes all requirements. I'm pushing up a tiny tweak to the test.

Issue #1831, opened 2021-03-09
Document Count on Data Group Tracking


Sakshi-Handa:
As a data curator
I would like to view the number of data documents in each data group on the /datagrouptracking/ page
So that I can prioritize workflows by the relative size of each data group

Definition of done:
• Applies to the /data_group_tracking/ page
• Add a 'document count' column for the number of Registered documents in a datagroup
Position to the right of 'Date Updated' column

PR #1838, opened 2021-03-16
1831 document count in data group tracking

mwfrost:
Closes: #1831 

This adds a column to the table in `http://127.0.0.1:8000/data_group_tracking/`


zachmunzenrider (changes requested):
Made small layout change, which Matt needs to approve or reject.

PR #1843, opened 2021-03-17
1833: improve cumulated products per puc sid performance

fang-yan-csra:
Used raw query to get cumulated product counts per puc/sid. Based on my testing, the query could take up to 8 seconds for big data set like DTXSID6026296 FO type.


michael-barbour (approved):
http://localhost:8000/chemical/DTXSID6026296/

This changes the page loading from a 75.9 second load time on my machine for production data "Water" to an 11.5 second load time.

mwfrost:
Merged off-cycle after discussion with EPA team and ad-hoc demo.

PR #1853, opened 2021-03-23
1829 chemical download

### LINKED ISSUE #1829: Download chemicals on /PUC/ page
As a factotum user
I would like to download chemical information related to a PUC
So that I can obtain data on all chemicals related to a specific PUC

Factotum users would like the ability to obtain all chemical information for a specific PUC. 
For example, if someone wanted all of the chemicals found in any Pet Shampoo product. 
This information is available on the PUC/# page, on the chemicals table. 
Ex. https://factotum.epa.gov/puc/259/

Definition of Done:
• Add a download button to the /puc/ page, next to the 'Chemicals' tab. 
• Available to logged in and guest users
• Ability to download as a .csv file with chemical information for the following fields:
○ DTXSID
○ Preferred Chemical Name
○ Preferred CAS

zachmunzenrider:
Closes #1829.

Working on this ticket, I realized the existing datatables download buttons weren't working. I tracked the issue down to an installation file that didn't include a component required for this functionality.

So, I ended up tweaking datasource_list and datasource_detail a bit, and then I also added download buttons to the other 2 tabs (product and document) on the PUC page. I also added a selenium test that will fail if these buttons ever go missing again. . . .


michael-barbour (changes requested):
The base functionality looks fine, but it appears with the AJAX loading that we are doing only the visible results are being added to the CSV.  IE if you're showing 10 entries your CSV will be 10 rows long despite the number of total results.

Tested on prod data with `puc/259/`.

** This was actually commented before but I didn't see it.  https://github.com/HumanExposure/factotum/issues/1829#issuecomment-805200566

michael-barbour (approved):
Tests are passing and the changes look good.  Tested on `puc/137/` on production with 2k results and download was basically immediate.

Issue #1785, opened 2021-02-01
Test dynamic filtering of multi-PUC products on Chemical detail page


Sakshi-Handa:
Test for a fix applied in ticket 1282.  The filtered JSON provided to the table should honor the uberpuc assignment logic.
1. Open http://127.0.0.1:8000/chemical/DTXSID7020637/ and click the funnel next to "Personal Care" in the PUC legend
2. Pick one of the products in the filtered table that looks like a lab supply, and open its Product detail page. In this case, http://127.0.0.1:8000/product/370826/
3. Change the PUC to "Lab supplies"
4. Return to the chemical page and switch the funnel filter to "Lab supplies"
5. Find the recently-changed product in the table

PR #1854, opened 2021-03-23
1785 Add test for PUC filtering on chemical detail page

mwfrost:
Closes: #1785 

This adds a test to confirm that the changes in #1282 corrected the Products shown when the user filters the chemical detail page on a PUC. It also includes a fix to the documents table on that page. 

The tables previously failed to respect the uberpuc assignments - when the AJAX query returned a set of products filtered by a PUC, the PUC was arbitrarily chosen from among all the PUCs assigned to the product. Now it uses the uberpuc selection logic. 

To reproduce the now-fixed error on the Documents table:

*On a branch other than this one,* and with the production data loaded, open the chemical detail page for Formaldehyde: `http://127.0.0.1:8000/chemical/DTXSID7020637/`

Click the filter button next to "Personal Care" to filter by that PUC. The number of Documents and Products will differ because some of the Documents are related to Products assigned to the Personal Care PUC where Personal Care is not the uberpuc.

Switch to this branch, reload the page, and the Documents that match those Products should disappear.

### DIFF
--- a/dashboard/fixtures/12_product_to_puc.yaml
+++ b/dashboard/fixtures/12_product_to_puc.yaml
@@ -158,17 +158,6 @@
     classification_method_id: MA
     classification_confidence: '1.000'
 - model: dashboard.producttopuc
-  pk: 20
-  fields:
-    created_at: 2019-05-10 14:45:56.307215
-    updated_at: 2019-05-10 14:45:56.307244
-    product: 878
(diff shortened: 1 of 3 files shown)


fang-yan-csra (approved):
Verified only uber PUC related products/documents show when filtered by PUC.

Issue #1804, opened 2021-02-23
Bulk Product Removal Pt 2. 


Sakshi-Handa:
Part 2 will add the product removal functionality. 

As a data curator
I would like the ability to remove large amounts of products from a given PUC 
So that I can remove miscategorized products and clean up PUCs

Definition of Done:
	• Applies to the bulk product removal page created in Pt 1. (#1802)
	• Ability to select multiple products from table for removal
	• Confirmation of action to 'Delete PUC Relationship' for selected products

PR #1855, opened 2021-03-23
1802 bulk product removal

### LINKED ISSUE #1802: Bulk Product Removal Pt. 1
Part 1 is for creating the new page and table. 

As a data curator
I would like to display and search all products within a PUC
So that I can find products within a PUC that are miscategorized

After implementing the 'automatic' (AU) PUC classification method, many products were placed in incorrect PUCs. I would like to be able to search for products within a PUC, and quickly remove products that don’t belong in that PUC. 
I would like this to be similar to the Attribute assignment tool, where curators search by PUC. https://factotum.epa.gov/bulk_product_tag/ 
Then have the ability to select multiple products to remove them. 

Comments:
Once you remove the product-to-PUC relationship, can you reassign using that same method?
Would like the table to only display products in this PUC table when they are UberPUC assignments. 
Would this remove all PUC assignments, or just the 'top layer' or UberPUC assignment?

Definition of Done:
	• New page for bulk removal of product-to-PUC assignments
	• Ability to search by PUC name 
	• Ability to search by product name within table
	• Table displays columns for: 
		○ Product name
		○ Classification method

michael-barbour:
closes #1802 

New page for authenticated users located at http://localhost:8000/bulk_remove_product_puc/ or through the nav bar  `Data Curation ->  Bulk Remove Products to PUC`.

- Has a DataTable with 3 columns.
- Only shows uber-pucs.
- Must be filtered by PUC (though this could easily be modified)
- May be filtered by Classification Method via checkboxes or icontains Product Title via search.
- Remove functionality to be added in #1804  

I left the select functionality and used a form template to help bridge the gap between this and the follow-up ticket.  They are currently untested as they are not actually used.


mwfrost (approved):
Works great. The checkbox filtering is a good solution.

PR #1863, opened 2021-03-24
1849: Improve data group tracking page loading time

fang-yan-csra:
Used a count query to get the document count for the group.


mwfrost (approved):
Looks good: now includes the performance benefits of both fewer queries and the `Count` function.

Issue #1852, opened 2021-03-22
Filter /Chemical/ page Functional Use Table based on PUC 


Sakshi-Handa:
As a Factotum user
I would like to filter functional uses based on PUCs selection on the /chemical/ page
So that I can see what functional uses a chemical would have in a given PUC.

The factotum userbase may want to know, what is the functional role of this chemical in this PUC? When I see Fluoride in Toothpaste, what is this chemical doing, what is the functional use?
So I would like the 'Functional Use' table to filter based on selected PUCs, just like the 'Document' and 'Products' Tables on the /chemical/ page. 

Definition of Done:
	• Applies to 'Functional Use' Table on /chemical/# page
	• Ability to filter functional uses displayed in table, based on PUC filter selection on bubble plot.

Issue #1830, opened 2021-03-09
 Add Table Dividers on Chemical Page


Sakshi-Handa:
Add Table Dividers on Chemical Page

As a factotum admin 
I would like to add table dividers to the tables on the /chemical/ page
So that we can compact the page and make it more user friendly. 

The functional use table hasn't been made yet (#1827). But when it is added, users will have to scroll a lot to view everything on this page. So I would like to add a table divider for each table. 

Definition of done:
• Applies to the /chemical/# page
• Add table dividers to the tables for Products, Documents, and Functional uses, similar to the /puc/ page
• Would like Products to appear first, as it is the most viewed table

Issue #1827, opened 2021-03-09
Functional Use Table on /chemical/ page


Sakshi-Handa:
As a Factotum user, I would like to view a functional use summary table on the /chemical/ page
So that I can see all information on functional uses related to a particular chemical 

Ex of chemical page.
 https://factotum.epa.gov/chemical/DTXSID0020232/

Definition of Done:
• Add a new table to the /chemical/# page for 'Functional Uses associated with [chemical]'
• Table Columns include:
	○ Document Type (can be Composition, Chemical presence, Functional Use)
	○ Associated Document (link to /datadocument/ page)
	○ Reported functional use
	○ Harmonized functional use (link to functional use detail page (created in #1825))

michael-barbour:
Added a comment to #1825 for last bullet point linking.  The link to the Functional Use detail page will be added when that ticket is completed.

Issue #1825, opened 2021-03-09
Functional Use Category Detail Page


Sakshi-Handa:
As a factotum user, 
I would like to view a summary detail page about a particular functional use category 
So I can see all information related to a particular functional use. 

Definition of Done:
•  Create a new detail page for each functional use category, similar to the /puc/ page. 
• Display the Functional Use Category Description on this page.
• Hyperlink from the /functional_use_categories/ Index page to each Detail page via the 'Title' 
• On the page, add 3 tables showing the Products, Documents, and Chemicals associated with this functional use category. 
• Please separate the tables with tabs.

1. Product Table (Some Composition docs have chemical ingredients with extracted functional uses)
	• Product Title (link to the /product/# page)
	• Chemical preferred name (the chemical within that product that has the harmonized functional use)

2. Document Table (data documents in which this func use category appears in a chemical record)
	• Data Group Type (since functional use data can be extracted from CO, CP, and FU documents)
	• Data Document in which the chemical record is found (link to data document) (link to /datadocument/ page)
	• reported functional use (the 'extracted' functional use reported on the data document)

3. Chemical Table (chemical records associated with this functional use category)
	• DTXSID (link to the /chemical/# page)
	• Chemical Preferred name 
	• Preferred CAS number

michael-barbour:
A new acceptance criteria is required for this ticket to prevent this from blocking #1827.

#1827 has a new table that links to this page through the "Harmonized Functional Use" category.  The render for this simply needs to change to an anchor link of whatever this table is.

Last bullet point in the acceptance criteria of #1827.

Sakshi-Handa:
On the Products Table, Instead of 'Chemical Name', add a column with the associated composition data document (and link to the /datadocument/ page). 
Also add a column for the PUC (display the uberPUC).

mwfrost:
Updated requirements for the three tables:

1. Product Table (Some Composition docs have chemical ingredients with extracted functional uses)
	- [ ] Product Title (link to the /product/# page)
	- [ ] Data Document title with link to data document detail page
	- [ ] Uber PUC (I have preemptively added a link to the PUC detail page)
	- [ ] Classification Method 

2. Document Table (data documents in which this func use category appears in a chemical record)
	- [ ] Data Group Type (since functional use data can be extracted from CO, CP, and FU documents)
	- [ ] Data Document in which the chemical record is found (link to data document) (link to /datadocument/ page)
[...]

PR #1864, opened 2021-03-26
Add table divider and functional uses table in chemical detail page

fang-yan-csra:
1. This PR covers ticket #1830, #1827 and #1852
2. Verify functional uses table is added to chemical detail page and can be filtered by PUC.

To Do: add link to functional use category page once #1825 is done.

Sample page:
http://localhost:8000/chemical/DTXSID6026296/

### DIFF
--- a/dashboard/fixtures/18_functional_use.yaml
+++ b/dashboard/fixtures/18_functional_use.yaml
@@ -83,13 +83,6 @@
     chem: 757
     report_funcuse: fragrance
 - model: dashboard.functionaluse
-  pk: 16
-  fields:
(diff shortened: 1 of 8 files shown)


michael-barbour (approved):
These changes look good and fit all of the acceptance criteria from #1830, #1827 and #1852.

I've added a comment to #1825 and #1827 about adding a link from the Harmonized Functional Use column to the Functional Use Detail page.  It makes more sense to get the completed code in rather than have this wait for the page to be done.

All tests are passing and the changes look and function well.

PR #1867, opened 2021-03-26
1804 bulk product delete action

### LINKED ISSUE #1804: Bulk Product Removal Pt 2.

michael-barbour:
closes #1804

Deletes the selected Uber PUC (per conversation with Product Owner).

One connection must be selected before submission or there will be an error.


fang-yan-csra (approved):
Verified that product to puc relationship can be deleted in bulk successfully.

PR #1870, opened 2021-03-29
1848 fix PUC admin

mwfrost:
The product count columns have been removed. There appear to be no tests that involved those values.


fang-yan-csra (approved):
Verified admin puc page can be accessed with no delay with production db.

PR #1871, opened 2021-03-29
1857 Chemical Download Label

### LINKED ISSUE #1857: Download Label for /chemical/ page
As Factotum admin
I would like to draw attention to the download button with a label
So users are aware that they can obtain information from the chemical page

Definition of Done:
- Applies to the /chemical/ page
- Add a label for "Download Products and Weight Fractions" to the download button added in #1814

michael-barbour:
closes #1857 

Just adds a label.  No test as this is a minor change, but I'd be happy to add one.


fang-yan-csra (approved):
Verified label added and download works as before.

Issue #1850, opened 2021-03-22
Remove unused PUC views and objects 


Sakshi-Handa:
Technical Debt Resolution

Definition of Done:
Clean up code, remove other solutions that were intended to solve the /chemical/ page issues

PR #1872, opened 2021-03-30
1850 PUC artifact cleanup

mwfrost:
Closes: #1850 

This removes a couple of the `django_db_views` models that I added in my search for a performant way to calculate products per PUC and chemical. I don't think any other artifacts can be removed.


fang-yan-csra (approved):
The unused ProductsPerPucAndSid and CumulativeProductsPerPucAndSid views are cleaned up.

PR #1873, opened 2021-03-30
1847: fix download pucs

fang-yan-csra:
Use cumulative_products_per_puc for products count.


mwfrost (approved):
Tested with prod data; cumulative totals match incremental counts.

Issue #1340, opened 2020-01-10
functional use model diagram


kdionisio:
Specify new data model for functional use relationships

Definition of Done:
Data diagram of proposed new data model structure

mwfrost:
This diagram shows the status quo inheritance relationship between ExtractedFunctionalUse and RawChem, which will be replaced in ticket 1322 with a one-to-many relationship.

![erd_before.png](https://images.zenhubusercontent.com/5be34a63f724c70c7221c972/96932d45-3c3e-4d3d-8d01-1dde6b2c96e9)

mwfrost:
![erd_before](https://user-images.githubusercontent.com/201835/72534557-e0884e80-382b-11ea-94f4-6457126b6e86.png)

The same image, but with an annotation asking about Composition data. 

We talked on the call about how every ExtractedChemical record can only have one `reported_funcuse` value, since the functional use is specific to the composition in which the chemical appears. But if we are extending the ExtractedFunctionalUse model in #1323 and #1326 with additional fields and curation metadata, it would make sense to enhance the functional use field for ExtractedChemicals in the same way.

debboutr:
@mwfrost here's what I was thinking for the new table created...
```
# this new setup will leave RawChem as the model that name/cas will be loaded into and
# then this table will have many funcuses that can be associated with the same chemical, one
# if the difficulties will be how to register. One way might be to have all the funcuses
# splittable in the CSV column by a given char, i.e. "," or ";" and the bulk_upload would create
# one record for each. This table name could stay ExtractedFunctionalUse I just changed it
# to show that it is a wholly new table.
class ReportFuncuse(CommonInfo): 
    
    chem = models.ForeignKey(
        "RawChem",
        on_delete=models.CASCADE,
        null=False,
        blank=False,
    )
    
    report_funcuse = models.CharField(
        "Reported functional use", max_length=255, null=True, blank=True
    )
```

debboutr:
the child model for the group type will change
```

mwfrost:
Right, the child model for a Functional Use data document would become RawChem because there's no corresponding model for a *chemical* that has been extracted for functional use analysis. Our current data model uses `ExtractedFunctionalUse` at this level in the data hierarchy, but that no longer makes sense. All the other models that inherit from `RawChem` do so because they add type-specific attributes to a chemical record. This series of tickets is all about relating multiple functional use objects to a single chemical record.

mwfrost:
[...]

PR #1874, opened 2021-04-01
1851 Sync Functional Uses

### LINKED ISSUE #1851: Synchronize Functional Use Categories with Curated Pairs
As a data curator, 
I would like a way to match reported uses to FU Categories
So that I can easily assign categories for known reported functional uses

I would like to create a process for easily assigning functional use categories to reported functional uses, based on the links that have already been established between these values. 
For example, if the reported use 'scent' has been linked to the category of 'Fragrance', when I add a new chemical record with a reported use of 'scent', I want to be able to connect it to 'fragrance'.

Is there a way that this can be done automatically? 

Definition of Done:
Ability to automatically update new reported functional uses to match previously established relationship between functional use and FU Category

michael-barbour:
closes #1851 

- [x] Data migration run on prod data

> Ability to automatically update new reported functional uses to match previously established relationship between functional use and FU Category

The "cleanest" way to do this is to set up a direct connection between a "Functional Use" (Reported Functional Use) and a "Functional Use Category" (Harmonized Functional Use).

The data migration should work both forward and backwards to bring prod data up to this new approach.

Caveats to this approach:
I've brought these up with Sakshi and she seems okay with them.
- Functional uses are not case sensitive due to mySQL not being case sensitive.
- Audit logs for functional uses do not currently link back to chemicals.  May need a ticket to address.
- API tests skipped due to being out of scope for OY5.  This will break some API endpoints.

### ERD
UML Diagram

### Links to some of the original ticket for reference
- First issue for multiple functional uses: https://github.com/HumanExposure/factotum/issues/1322
- First diagram for multiple functional use: https://github.com/HumanExposure/factotum/issues/1340#issuecomment-575800532


michael-barbour:
Fang's changes to the audit log look good.

fang-yan-csra (approved):
I'm able to migrate/rollback data with latest production dump. Verified to add/remove reported functional use from UI. Also Admin pages can add functional use, link chemicals as well.

PR #1876, opened 2021-04-02
1832: Display Raw Composition data on Chemical Card

fang-yan-csra:
Include raw min/central/max composition data and weight fraction lower/central/upper analysis data on composition chemical card.

Sample page: http://localhost:8000/datadocument/156051/


mwfrost:
Other sample pages to use when demonstrating the ticket:
http://127.0.0.1:8000/qa/extractedtext/267812/ has lower and upper, http://127.0.0.1:8000/qa/extractedtext/268248/ has central

mwfrost (approved):
Tested with prod data on cards with central and min/max ranges.

PR #1881, opened 2021-04-06
1858: Download Products and Weight Fractions Related to a PUC

fang-yan-csra:
Added download button on puc detail page to download composition chemicals related to puc.

To test, go to any /puc/{id} page, click the download products and weight fractions button. Verify the csv file including all chemicals related to this puc is downloaded.


mwfrost (approved):
All specified fields are in the output file, the download is prompt even with production data,

PR #1889, opened 2021-04-09
1821: Download chemicals for CO docs

fang-yan-csra:
Add download chemical data button for CO type of data document.
Example page: http://localhost:8000/datadocument/7/


michael-barbour (approved):
Changes look good.  Tested on February prod data http://localhost:8000/datadocument/1490771/ with 153 chemicals (largest set) and the download is instant.  All data seems correct.

PR #1890, opened 2021-04-09
1880: fix functional use category grid style

fang-yan-csra:
The id=dataGrid ag-grid had a fixed height 500px style. Changed to use a fix-height-grid class for the affected /functional_use_curation/ page.
Test: verify the grid on /functional_use_curation/  and /functional_use_categories/ page looks fine.


michael-barbour (approved):
http://localhost:8000/functional_use_categories/ looks great.  No issues with overflow.

PR #1894, opened 2021-04-13
1869: update django rest framework version

fang-yan-csra:
Tested factotum API still run. All test cases are skipped now.


mwfrost (approved):
The API still launches, but several of the endpoints are broken due to changes made to the data model. The tests have not been updated.

PR #1895, opened 2021-04-14
1868: update PyYAML version

fang-yan-csra:
Used `pip install --ignore-installed -r requirements.txt` to update the PyYAML package. Without the --ignore-installed option, pip install errors out with "ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall."


mwfrost (approved):
I updated using the command specified in the PR and ran some tests to make sure the YAML still loads.

Issue #1614, opened 2020-08-25
display of list presence tag sets on list presence tag summary page


kdionisio:
As a Factotum user
I want to see the list presence sets associated with a document
So that I know which documents are necessary for my needs

In #1756, we added a display of the unique CP tags associated with a data document. That accounts for all list presence tags that appear in all chemicals in the document, and displays them once on the sidebar. 
Ex. https://factotum.epa.gov/datadocument/1496321/
I would like to display this same 'unique tag keyword set' on the /list_presence_tag/# page, on the Documents table. So that users know what tags are present within that data document. https://factotum.epa.gov/list_presence_tag/457/ 

Definition of Done:
- On the /list_presence_tag/# page, in the 'Documents' table, add a second column to the table which displays the list presence tag keywords associated with the document. 
- The unique list presence tag keywords should be displayed, as found on the /datadocument/# page for CP documents. 
- User should have the ability to filter/subset this new second column of the table by a keyword or list presence tag (can search for a keyword/tag)

kdionisio:
potentially split into two tickets; also hold and rewrite - because each chemical record within a document may have it's own list presence tag set, this request will cause issues

Sakshi-Handa:
I am refining this ticket based on what has been done in #1756. Where we added the 'unique tag set' to the /datadocument/ page for CP documents.

mwfrost:
@Sakshi-Handa I was unable to make this full set of results searchable without slowing down the load time an unacceptable amount. Do you mind deferring the search function into its own ticket?

PR #1896, opened 2021-04-14
1614 Show List Presence Tags per Documentes: 

mwfrost:
Closes: #1614 

The same way the data document detail page shows the unique list presence tags that occur across all the chemicals in it, this branch adds a hyperlinked set of list presence tags to the document's row on the List Presence Tag Index page. Use `/list_presence_tag/457/` for an example. 

When reviewing, please evaluate the performance with prod data and suggest any opportunities to optimize the query.


dashboard/tests/integration/test_list_presence_tag_page.py
```diff
@@ -70,9 +70,15 @@ def test_list_presence_tag_page(self):
...
         wait.until(
             expected_conditions.text_to_be_present_in_element(
                 (By.ID, "document_table_info"), document_string
             )
         )
+        # Check for tag list next to document
+        tagset_cell = self.browser.find_element_by_xpath(
+            "//*[@id='document_table']/tbody/tr/td[2]"
+        )
+        print(tagset_cell.text)
```
michael-barbour, line 84:
```suggestion
```

dashboard/views/ajax.py
```diff
@@ -158,17 +159,34 @@ def get(self, request, *args, **kwargs):
...
+            tag_names = []
+            for tag in sorted(tagset, key=lambda o: o.name.lower()):
+                tag_names.append(
+                    f"<a href='{reverse('lp_tag_detail', args=[tag.id])}' title='{tag.definition or 'No Definition'}'>{tag.name}</a>"
+                )
+            tag_links = " ; ".join(tag_names)
+            linked_title = f"<a href='{reverse('data_document', args=[cpcat.data_document.id])}' title={cpcat}'>{cpcat}</a>"
+            json_data.append([linked_title, tag_links])
+
+        return JsonResponse({"data": json_data}, safe=False)
```
michael-barbour, line 189: > When reviewing, please evaluate the performance with prod data and suggest any opportunities to optimize the query.

There is quite a hit when changing the loading of data from paginated to not-paginated data.  This new approach seems to require a query for each data document that contains the tag before the table will show any data.

Compare against adding a specific render function.

```python
class ListPresenceDocumentsJson(FilterDatatableView):
    model = DataDocument
    columns = ["title", "tags"]

    def render_column(self, row, column):
        if column == "tags":
            tag_str_list = []
            tags = ExtractedListPresenceTag.objects.filter(
                extractedlistpresence__extracted_text__data_document=row.id
            ).order_by("name").distinct()
            for tag in tags:
                tag_str_list.append(f"<a href='{reverse('lp_tag_detail', args=[tag.id])}' title='{tag.definition or 'No Definition'}'>{tag.name.lower()}</a>")
            return " ; ".join(tag_str_list)
        return super().render_column(row, column)

    def get_initial_queryset(self):
        qs = self.model.objects.filter(
            extractedtext__rawchem__extractedlistpresence__tags__pk=self.kwargs[
                "tag_pk"
            ]
        ).distinct()
        return qs
```
While sorting and searching will be unavailable by default and without some work, the initial data load will be 1 query to fetch the documents and a second for each visible row to fetch the assigned tags.  If 10 rows are shown it will be 11 total queries instead of all documents with a connection to that tag + 1.

michael-barbour: I'll put this up on `1614_lp_tagsets_paginated` if you want to check it out.

mwfrost: I made these changes directly on my branch without merging the `_paginated` branch. Thanks for writing it up.

michael-barbour (approved):
Tests run and all data seems correct.

PR #1897, opened 2021-04-14
Sprint release Newt

kbfepa:


kbfepa (commented):
Notes:

Migrations noted on:
- `dashboard/migrations/0186_remove_views.py`
- `dashboard/migrations/0187_add_functionalusetorawchem_model.py`

Packages:
- changes in `requirements.txt` with the upgrades of `djangorestframework` and `drf-yasg` from security update.

Additionally, following noted by Michael Barbour:
- Data migration expected to take 5-20 minutes; Making a db back-up and possibly running the migration on a cloned db recommended; Data migration is wrapped in a transaction so data shouldn't be at risk.

PR #1898, opened 2021-04-15
1879: Add Count to Chemical table on PUC page

fang-yan-csra:
Added count of Chemical table on PUC page and download.


dashboard/views/ajax.py
```diff
@@ -107,25 +107,24 @@ def filter_queryset(self, qs):
...
-    columns = ["sid", "true_cas", "true_chemname"]
+    columns = ["sid", "true_cas", "true_chemname", "raw_count"]
 
     def get_initial_queryset(self):
         qs = super().get_initial_queryset()
         puc = self.request.GET.get("puc")
         if puc:
-            vals = (
-                RawChem.objects.filter(dsstox__isnull=False)
-                .filter(Q(extracted_text__data_document__products__puc=puc))
```
mwfrost, line 118: This should only show uberpuc assignments

mwfrost (changes requested):
The count should observe the UberPUC logic. 

In the most recent prod data, Product 582303 has been assigned to PUCs 77 and 309, with 309 as the uberpuc. Because of the duplicate assignment, the chemicals table on the detail page for PUC 77 includes DTXSID00100094 when it should not.


URLs and queries for checking:
```
http://127.0.0.1:8000/chemical/DTXSID00100094/
http://127.0.0.1:8000/product/582303/

select * from dashboard_producttopuc where product_id = 582303;
select * from dashboard_rawchem where id = 4715367;
```

The query that needs to be revised is here:
https://github.com/HumanExposure/factotum/pull/1898/files#r616792027

dashboard/views/ajax.py
fang-yan-csra: Pushed to apply uber puc logic

mwfrost (approved):
Tested with production data, works correctly.

PR #1901, opened 2021-04-20
1878 funcuse curation page

### LINKED ISSUE #1878: Create Functional Use Curation Detail page
As a Factotum Admin
I would like to view and access the data document associated with a functional use record
So that I can track and edit individual reported functional uses.

We need the ability to get back to the original data documents and edit the reported functional uses, so we can separate out lists into single functional uses. 

Definition of done:
• Applies to https://factotum.epa.gov/functional_use_curation/
• from the index page, add a link to the 'count' leading to the functional use detail page. 
• Page lists all chemical records with that reported functional use, and links back to /datadocument/ page

michael-barbour:
closes #1878 

Adds a new page and some tests for rendering.  The chemical name is a case which renders either the `dsstox.true_chemname` if one exists or the `raw_chem_name` if one does not.

Fixed RawChemFactory to correctly use the `is_curated` param correctly, and have `ExtractedCompositionFactory.create(functional_uses=[])` take an iterable instead of taking a number.


fang-yan-csra (approved):
Verified that count links to new functional use curation detail page, in which chemical links to data document page as requested.

Issue #1882, opened 2021-04-06
Functional Use Table on /PUC/ page


Sakshi-Handa:
As a factotum user,
I would like to see a table of functional uses on the /PUC/ page
So I can view all functional uses related to a particular PUC

We would like to add a table to view the functional uses associated with Composition type documents with products linked to a particular PUC. Similar to how the Functional Use table on the /chemical/ page can be filtered by PUC. Instead of searching by a single chemical, we want to see functional uses for all chemicals within that PUC.

Definition of done:
• Add a new table to the /PUC/# page, within the table dividers
• The Tab heading is "Functional Uses"
• Within the table, add columns for:
	• Data Document - link to data document page
	• Preferred Chemical Name
	• Preferred CAS
	• Reported Functional Use
	• Harmonized Functional Use

mwfrost:
@Sakshi-Handa Do you want any distinction made between curated and non-curated chemicals? The "preferred" chemical name and CAS are both pulled from the DSSTOX record when one has been curated onto the Raw Chemical record. Should the user know whether the row in this table is showing a true or a raw CAS or chemical name?

PR #1903, opened 2021-04-22
1882 Functional use table on PUC page

mwfrost:
closes: #1882


fang-yan-csra (approved):
Verified Functional Uses table works fine on PUC page.

PR #1904, opened 2021-04-23
1861: Habits and Practices Table on PUC page

fang-yan-csra:
Use http://localhost:8000/puc/2/ for sample db testing.


michael-barbour (approved):
Changes look good and all tests pass

Issue #1885, opened 2021-04-07
Manual Composition QA  Pt 1 (Main Index page)


Sakshi-Handa:
As a data curator
I would like to view all Composition Data Groups that contain manually extracted documents 
So that I can determine the status of Data Groups and access QA. 

There is currently no QA workflow for Manually extracted Composition documents. 
Documents that are manually extracted automatically get linked to the 'manual dummy' extraction script. 
I would like the manual QA to be similar to Chemical Presence QA, in that it is organized by Data Group. We would QA all of the documents within each Data Group that were extracted manually (linked to the dummy_manual extraction script) https://github.com/HumanExposure/data_mgmt_scripts/blob/master/dummy_Manual

We would NOT need to make a QA Group to check a certain percentage of the documents. We would like to display ALL manually extracted documents within a Data Group, as errors can appear in any manually extracted document. 

Definition of Done:

- [ ] Add a new tab under 'QA' for 'Composition - Manual'
- [ ] Create new page for Manual Composition QA, based on Data Groups (similar to chemical presence QA, /qa/chemicalpresence/  )
- [ ] Include the following columns
	- [ ] Data Group (should only show Composition groups that have manually extracted documents)
	- [ ] Document Count (number of Manually Extracted documents in Data Group)
	- [ ] Percent QA Checked (percent of documents checked in Data Group)
	- [ ] QA Summary (QA Summary Page will be created in pt. 3)
	- [ ] QA Status (QA Group page will be created in pt. 2)

PR #1912, opened 2021-04-30
1885 QA index page for manually extracted composition docs

mwfrost:
closes: #1885 

The seed data does not include manually-extracted composition records, so I used a factory to create them in my test. If you want to add them to your local dev environment, this code from the test should work in the shell:

```{python}
from dashboard.tests import factories

from dashboard.models import (
    ExtractedText,
    Script,
    DataGroup,
)
script = Script.objects.filter(title="Manual (dummy)", script_type="EX").first()
dgs = DataGroup.objects.filter(group_type__code="CO")

for dg in dgs:
    # add some manually-extracted records in proportion to
    # the count of records
    doc_count = ExtractedText.objects.filter(
        data_document__data_group=dg
    ).count()
    factory_count = (doc_count // 2) + 1

    extext = factories.ExtractedTextFactory.create_batch(
        factory_count,
        data_document__data_group=dg,
        extraction_script=script,
        data_document__title=f"{dg.name} manually extracted doc",
    )
```

### DIFF
--- a/dashboard/forms/forms.py
+++ b/dashboard/forms/forms.py
@@ -262,7 +262,10 @@ class ExtractedTextHPForm(ExtractedTextForm):
 
     def __init__(self, *args, **kwargs):
         super(ExtractedTextHPForm, self).__init__(*args, **kwargs)
-        self.fields["extraction_completed"].widget.attrs.update({"class": "ml-2 align-middle form-control-sm"})
+        self.fields["extraction_completed"].widget.attrs.update(
+            {"class": "ml-2 align-middle form-control-sm"}
+        )
+
 
 class ExtractedTextFUForm(ExtractedTextForm):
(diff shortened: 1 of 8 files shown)


fang-yan-csra (commented):
data looks good.
I would suggest make the table datatables like other pages, with page/search/sort etc
Also, the original AC does not include URL column, not sure if we’d add it

mwfrost:
> Also, the original AC does not include URL column, not sure if we’d add it

The original composition QA index is organized by Extraction Script. The ticket asks for this page to be like the Chemical Presence QA page, which is organized by Data Group. The Data Group URL is provided in that table, so I think we should include it here.

fang-yan-csra (approved):
Verified /qa/manualcomposition/ page displays data as requested

PR #1914, opened 2021-05-04
1773 paginate datadocument detail

### LINKED ISSUE #1773: Pagination of Chemical Presence Data Documents
As a factotum user
I would like to improve the performance of data document pages
So that I can quickly navigate through chemicals of interest on pages with many chemicals. 

For some data documents with large amounts of chemical records, it takes some time to load the page, and is slow to respond to user input. Most data documents have less than 10 chemicals, so we don’t run into this often. But for the documents with 1000+ chemicals, page performance is impacted. 

Ex. https://factotum.epa.gov/datadocument/1557970/
This chemical presence document has 6000+ chemicals. It takes over a minute to load the page, and it is slow to select the chemical cards and assign keywords. 

Definition of done:
	•Paginate  /datadocument/# page for chemical list presence documents to return 500 chemicals at a time.

zachmunzenrider:
Closes #1773.  As per ticket, the page size is 500 cards, which isn't testable with our seed data. With prod data you can use http://127.0.0.1:8000/datadocument/1557970/ to test. It should be barely usable on the sprint-branch, but work fine on this branch.


fang-yan-csra (approved):
Verified pagination works fine for large set of chemicals on data document page. QA chem cards work fine as well.

PR #1924, opened 2021-05-06
download func uses button in chem detail page

BryanMG957:


michael-barbour (approved):
Looks good.  Tests pass

PR #1925, opened 2021-05-06
1862 provisional dtxsid assignment

### LINKED ISSUE #1862: Provisional DTXSID Assignment
Conduct research into dynamic assignment of provisional DTXSIDs

As a factotum Admin
I would like to add provisional DTXSID assignments to uncurated chemical records
So that i can view these chemicals within factotum

definition of done:
- Ability to assign provisional DTXSIDs for raw chemical names that are an exact match to an existing 'Raw Name and CAS' pair that have already been assigned to a DTXSID, True Name and CAS. 
- There should be no conflicts between DTXSIDs assigned to the same raw name/CAS pair. 
- Add a 'provisional' flag to provisionally assigned DTXSIDs

michael-barbour:
closes #1862 

Added new [.env variable](https://github.com/HumanExposure/factotum/wiki/Configuration#provisional_assignment_schedule-toc) `PROVISIONAL_ASSIGNMENT_SCHEDULE` with a default to 3:00am nightly.  The schedule accepts a space separated [crontab(5)-like](https://man7.org/linux/man-pages/man5/crontab.5.html) string.             

```
field          allowed values
-----          --------------
minute         0-59
hour           0-23
day of month   1-31
month          1-12 (or names, see below)
day of week    0-7 (0 or 7 is Sunday, or use names)
```

Added a migration with a mysql specific [prefix index](https://dev.mysql.com/doc/refman/8.0/en/column-indexes.html#column-indexes-prefix) to migration dashboard 0190.  This speeds up the speed of provisioning by quite a bit (~50x).

On my build of production, I'm getting 619,859 new associations when I provision and 1,322,802 raw chems with no dsstox that have nothing to associate with.

## Production
Not much was needed for building prod `docker/docker-compose.yaml;docker/docker-compose.override.yaml` to work.  The `celery-beat` container already builds from factotum.  You'll need to this branch to test it.

## Development
~Attempts to add a background beat process to django runserver.  This is loud and might hang so it might be worth building in environment variable to opt-out.  It may also be worth building celery and celery beat into the default docker-compose.yaml file.~

This was giving issues to our windows devs so it was reverted.  Instead a beat process will need to be run in another window. `celery beat -A factotum`


michael-barbour:
Clarification from Sakshi on the "no conflicts" bullet point:

> So 'conflicts' in the sense Kristin was talking about were when a name/cas pair has been assigned to 2 or more different DTXSIDs.
> We have been sending chemicals over to DSSTOX for the last 2-3 years. and over time, they might have updated their chemical mappings. But we don't get the updates. So for example, a raw chemical/CAS we sent in 2019 might have gotten DTXSID A. then if we sent that same chemical/CAS pair in 2021, it would be assigned DTXSID B.
> So this ticket was saying, if there are multiple DTXSIDs for the same raw chemical/CAS pair, then don't assign a provisional DTXSID.
> In theory, we would want to assign the 'newer' mapping. But I think Kristin's concern was if there was a conflict between them, that we can't be sure which one is 'correct'

michael-barbour:
Moved wrong issue on pipeline.  Reopening to be merged.

PR #1930, opened 2021-05-12
Hotfix Provisional DTXSID Assignment

michael-barbour:
There was a bug where if more than one dsstox id was returned found, the record would not be provisioned.
A `.distinct()` was added to change this to requiring more than one different dsstox id.

Also dropped the auditlog extracted_text_id foreign key constraint. There are rows in production rawchem table whose extracted_text_id that does not exists in extractedtext table, which is wrong. This caused error throw when set the provisional sid. Drop the constraint is auditlog is fine as this is audit trail anyway. But how the production database get into this state is still unknown/unresolved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,6 +106,7 @@ target/
 # celery beat schedule file
 celerybeat-schedule
 celerybeat-schedule.*
+celerybeat.pid
 
 # SageMath parsed files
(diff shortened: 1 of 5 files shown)


mwfrost (commented):
Tested with previous exceptions and approved

PR #1932, opened 2021-05-14
1916 comments modal

### LINKED ISSUE #1916: Submit Comments on a Factotum Page
As a Factotum user
I would like the ability to submit comments and questions related to a particular page in Factotum
So that I can get help with the information I am viewing. 

Currently, we have a general feedback page (https://factotum.epa.gov/feedback/)
But we would like it to be more immediately available for users, so we can get their feedback as they navigate the site. 
Maybe this would mean reworking the feedback page to be a pop-up form? Or open in a new tab? So the user can provide comments related to a particular document or chemical.

Definition of Done:
	• Relocate the Comments page to be clearly visible as a tab on the Main Toolbar. 
	• Add hover text to the Comments button: "Provide feedback on this page"
	• Available for logged in and guest users
	• Add text to the comments page: "Use this form to provide feedback on pages, report errors, suggest data sources, ask questions"
	• On the Feedback page, capture and store the user's current page url.
        • On the Feedback page, add a Subject field.

michael-barbour:
closes #1916 

Converts the comment create page into a comment create modal that responds to AJAX.

This does not capture the full url but only the uri.

This is the first time an Anonymous user could make a change to a "CommonInfo" field so I've added a check to account for that.


fang-yan-csra (approved):
Verified comments available for all pages. User input can be saved to db successfully.

PR #1935, opened 2021-05-18
1917 add columns to products downloads

BryanMG957:


fang-yan-csra (approved):
Verified document date and classification method included in downloaded csv.

Issue #1939, opened 2021-05-18
Statistics on Data Source page


Sakshi-Handa:
On the data source page, the % Registered and % Uploaded columns can be greater than 100%. 
Ex.  https://factotum.epa.gov/datasource/98/

I believe this field is based off of the number of 'Estimated Records'. But we could base it on what is really in Factotum. 
What is the difference between registered and uploaded?
Maybe we should change one to 'extracted'.

Definition of Done:
	• On /datasource/# page, % Matched and % Extracted fields represent a percentage out of 100. 
		○ Percentage of documents matched out of total registered to the data source
		○ Percentage of documents extracted, out of total registered to the data source

PR #1950, opened 2021-05-25
1939 fix data source percentage stats

mwfrost:
closes: #1939 

The percentages shown on the form previously used the estimated record counts, which resulted in meaningless and weird summary numbers. Now the "% Uploaded" and "% Extracted" percentages show the share of all documents in a data source that have been associated with a file and with extracted text data, respectively.


fang-yan-csra (approved):
I pushed to handle divide by 0 error. The percentages are calculated as requested

PR #1953, opened 2021-05-25
1920 composition data

### LINKED ISSUE #1920: Display N/A for Documents with No Reported Composition Data
As a data curator, 
I would like the ability to designate "Not applicable" on the reported field of Composition type data documents
So that I can convey that there is no composition information available for that document. 

In the data downloads for Product and Weight Fractions, there are many documents that have no composition data present in the reported and analyzed weight fraction fields. This is in many cases, because the documents themselves don’t provide composition information. But when users see the data, they think the fields are blank because we haven't curated the data yet. So we would like a way to convey to the end user that the data has been extracted, but there was no composition information to extract. So a way to display 'not applicable' on the reported composition fields. 

Ex. Ingredient Disclosure documents don’t have composition data. 
https://factotum.epa.gov/datadocument/1385390/

Documents with Composition data
https://factotum.epa.gov/datadocument/1553495/

Thoughts:
Maybe we can use the "unit type" field for this? Or raw central composition.
Would like to be able to set at the data group level in addition to the document level. Certain data group types don’t have composition data, i.e. Ingredient disclosures. Or from certain sources, like Graham Chemical.  
But want it to be overridden when there is data at the document level. If there is composition data, don’t hide it. 

Definition of Done:
	• Ability to convey on /datadocument/ page and data downloads that certain Composition type documents have  no composition data available.

michael-barbour:
closes #1920 

Adds a new field `ExtractedComposition.has_composition_data` and adds this field to the form for modification.  Chemicals can be saved with this field false to show as "No Composition Data".

Added some JS to the form to disable the raw composition data fields on the add/edit modals.

Fixed a bug with the HP page that was causing a test to fail.


zachmunzenrider (approved):
The has_composition_data field is a clean solution for this ticket. UI works well, tests all pass. 

The ticket mentions being able to set this at the datagroup level, but based on the Definition of Done, I think that implementation must be intended for a separate ticket.

Issue #1938, opened 2021-05-18
Update Factotum APIs


Sakshi-Handa:
As a Factotum user
I would like to ensure that Factotum APIs are maintained
So that I can still obtain data via API

Definition of Done:
Update Factotum APIs that were broken during data model updates (such as PUCs, Products, Functional Use)

mwfrost:
GDIT staff: please use the O&M timecode for this.

PR #1956, opened 2021-05-27
1938 fix API

mwfrost:
Closes: #1938 

This was suspiciously easy. Where the `FunctionalUse` endpoint had previously used the `FunctionalUse` model, I changed the serializer and the view to use `FunctionalUseToRawChem` instead. This approach seems to better suit the user's understanding of "functional use" referring to the combination of a chemical and a reported functional use string.

Once I made those changes, all the tests passed.


fang-yan-csra (approved):
Looks good. All API tests passed.

PR #1957, opened 2021-05-27
1945 pmid register record

### LINKED ISSUE #1945: Add PubMed ID to register records CSV
As a factotum admin
I would like to add the PubMed ID as a field to the register records csv upload
So that I can capture this information for the documents I upload

The register records csv is required when creating a new data group and defining the files that will be added to the group. 
https://factotum.epa.gov/datasource/98/datagroup_new/

PMID already exists as a field in the Literature Monitoring data type. 
https://factotum.epa.gov/datadocument/edit/1555997/

Definition of Done:
	• Applies to the Register Records csv upload, when creating a new data group.
	• Add PubMed ID as an optional field to the Register records csv upload.

michael-barbour:
closes #1945

Very straight forward.  The test the functionality is working is on `test_csv_upload_row_verification`.

This field is currently only applicable on Literature Monitoring documents and will only show on the edit form for LM documents.
Rather than write a whole bunch of business logic around the LM group in particular, it would be easier if we just left it blank on data documents where it was not applicable.

```
filename,title,document_type,url,organization,subtitle,epa_reg_number,pmid
Cal_Pesticide_Residues_1987.csv,Example Datadocument from CSV,UN,http://www.epa.gov,org,cpr,,12345
```


fang-yan-csra (approved):
Verified pmid field in csv upload when create a new data group and saved to db successfully.

PR #1977, opened 2021-06-07
reported_funcuse cleanup

### LINKED ISSUE #1942: Clean up reported functional uses
As a factotum admin
I would like to clean up reported functional use entries that have leading or trailing spaces
So that I can simplify and reduce rows on the functional use curation table

Definition of done:
	• Remove leading and trailing spaces from 'reported functional use' entries, so that they appear in one row on the /functional_use_curation/ page

zachmunzenrider:
Closes #1942. Tested successfully against the latest prod dump. I don't believe it will do anything to hurt dev data.


fang-yan-csra:
I pushed to use code to fix data. Also tested reported functional use is saved without leading/trailing spaces, from csv upload or form.

mwfrost:
Fang's revised migration worked fine on the latest production SQL dump. There are no orphaned records in the join table:
```
SELECT * from dashboard_functionalusetorawchem where NOT functional_use_id IN(select id from dashboard_functionaluse);
0 row(s) returned
```

mwfrost (approved):
Approved after testing with production data.

Issue #1962, opened 2021-06-01
Functional Use Data Document Download


Sakshi-Handa:
As a Factotum user 
I would like to the download chemical data for functional use documents. 
So that I can view and use this data

Ex. https://factotum.epa.gov/datadocument/1512907/
There is currently no download ability for "Functional Use" type documents. 

Definition of Done:
• Applies to the data document page for Functional Use documents
• Ability to download chemical data via the /datadocument/ page, similar to downloads for CO and LP documents
• Available to logged-in and guest users
• Add the following columns to the csv download:
	• Data Document Title	
	• Data Document Subtitle	
	• Organization	
	• Raw chemical name	
	• Raw CAS	
	• DTXSID	
	• True Chemical Name	
	• True CAS
	• Provisional Flag
	• Reported Functional Use
Harmonized Functional Use

Sakshi-Handa:
display in one 'functional use' field
reported: abrasive harmonized: Abrasive | (next entry)

PR #1984, opened 2021-06-14
Bulk LP Download

BryanMG957:


zachmunzenrider (changes requested):
This works well, but it's missing a test. You can probably grab and tweak Fang's test from #1962.

zachmunzenrider (approved):
I'm adding a test in my forthcoming branch

PR #1993, opened 2021-06-18
add url to admin comment page

BryanMG957:


mwfrost (approved):
URL is now visible - he URLs stored in the existing prod database show up on the admin page.

PR #1994, opened 2021-06-18
add true chem name/cas to curated chem detail page

BryanMG957:


mwfrost (approved):
HTML tags had been copied and pasted without updating IDs and the test had not been updated, but with that done this is approved.

PR #1995, opened 2021-06-21
1987 Unassign reported functional use

### LINKED ISSUE #1987: Bulk Remove Reported Functional Uses
As a data curator
I would like a way to easily delete all instances of a functional use via the functional use curation detail page
So that I can delete information in the 'reported functional use' field that is not a functional use

Ex. https://factotum.epa.gov/functional_use_curation/2043/
The word 'personal care' has been stored in the reported functional use field for 78 chemical entries. I would like a way to quickly remove all instances of 'personal care' wherever they appear on the chemical records, through a delete button on the functional use curation detail page. Then we can use the "clean up" button on the main page to remove the entry from the table. 

Definition of Done:
• Add the delete functionality to the functional use detail page /functional_use_curation/#/
• Delete button would remove all instances of a reported functional use from chemical cards on the /datadocument/ page
• Required a confirmation of delete action

mwfrost:
Closes #1987 

To test, start on `http://127.0.0.1:8000/functional_use_curation/` and choose one of the combinations that should be unassigned. Click the hyperlinked count for that combination, which opens the detail page, like `http://127.0.0.1:8000/functional_use_curation/9583/`. Click the "Unassign..." button and confirm to delete the join records.


zachmunzenrider (approved):
Functionality looks good, code is clean. I'm not loving the orange buttons that have snuck onto this page, and I do prefer the modals we often use for confirmation messages, but those aren't important enough to hold up the ticket. Approved!

Issue #1964, opened 2021-06-01
Bulk Download Composition Data


Sakshi-Handa:
As a Factotum Admin
I would like to make available a bulk download of all Composition related data on Factotum
So that users can get up-to-date data via the application

Definition of Done:
• Add a button to 'Bulk Download Consumer Product Data' on the /get_data/ page 
• Allows users to download a csv file of Composition data, including the following fields
		○ Data Source
		○ Data Document Title
		○ Document Date 
		○ Product Name 
		○ uber PUC (General Category	Product Family	Product Type)
		○ PUC Classification Method
		○ Raw Chemical name
		○ Raw CAS
		○ DTXSID
		○ Curated Chemical name
		○ Curated CAS
		○ Provisional Chemical Curation Flag
		○ Raw Min Comp	
		○ Raw Max Comp	
		○ Raw Central Comp	
		○ Unit Type	
		○ Lower Weight Fraction	
		○ Upper Weight Fraction	
		○ Central Weight Fraction	
		○ Weight Fraction Type

mwfrost:
The output file needs to include the PUC columns

mwfrost:
The PUC Kind should be included as another column, and the three PUC levels should be provided as separate columns.

PR #1997, opened 2021-06-23
1964 fix 

mwfrost:
Closes: #1964 , which was re-opened during sprint review. During review we realized that the branch did not include the PUC-related columns that the ticket required. They have been added, and the test has been updated.


zachmunzenrider (approved):
Looks great. Thanks for taking this!

PR #2000, opened 2021-06-29
1886 manual composition qa page

BryanMG957:


fang-yan-csra (approved):
Manual composition scripts QA page works as requested.

PR #2014, opened 2021-07-01
Add 'QA Approved' column to summary pages

BryanMG957:


fang-yan-csra (approved):
Verified QA Approved column displays correct data on all summary tables.

PR #2015, opened 2021-07-02
1992: set no composition data for data group

### LINKED ISSUE #1992: Setting N/A for No Composition Data Documents
This ticket is related to #1920

I don't think the No Composition Data flag is set by the data group tracking page, as we decided on in the previous ticket. It looks like there is no indicator on the documents within the "N/A" data groups, like ingredient disclosures. https://factotum.epa.gov/datadocument/1516336/
And in the data download for this document, the "Has Composition data" column shows "TRUE" for all chemicals. When they should say False, because there is no composition data present. 

The 'has composition data' flag on the chemical edit form is working, and displays on the chemical card. The data download updates as well. 

Maybe we need to add the flag to the /datagroup/ page, and curators can manually select it for groups that have no composition data available. 

Definition of Done:
- Ability to set the "Has Composition Data" flag at the data group level, via the /data_group_tracking/ page

fang-yan-csra:
Closes #1992 

Added "No Composition Data" button on data_group_tracking edit page for CO type. When clicked, will confirm user first, then reset has_composition_data flag for all related extractedcomposition chemicals and the Composition Clean step status to  N/A.


mwfrost (approved):
Tested successfully with Data Group 190. Looks good!

PR #2016, opened 2021-07-02
2010 change get data page layout

BryanMG957:


fang-yan-csra (approved):
Get data page layout changed as requested.

PR #2017, opened 2021-07-06
2006 add description harmonized medium

BryanMG957:


fang-yan-csra (approved):
Verified migrate works and can be rolled back. Admin can add description.

PR #2022, opened 2021-07-07
1983: fix style and extend slide interval

fang-yan-csra:
1. fixed /news/ page style
2. set the carousel time interval to 5 seconds


mwfrost (approved):
Layout looks much better, URL is consistent with others.

Issue #2023, opened 2021-07-07
Error Saving LM Card Edits


Sakshi-Handa:
There is an issue saving edits to the LM chemical card via the /datadocument/ page. 
After adding data to fields in chemical edit form and clicking 'save edits', the entered data is not stored in the edit form and does not appear on the chemical card. 

Ex. LM document
https://factotum.epa.gov/datadocument/1555997/#chem-card-5156463

Definition of Done:
applies to chemical edit form and chemical card on /datadocument/ page for LM documents
ability to manually add and save data through the chemical card edit form, which also appears on the chemical card.

mwfrost:
Somehow, when these two datagroups were populated with their extracted chemical records, there were duplicate `ExtractedComposition` objects created. The code that should update the `ExtractedLMRec` object is instead trying to save changes to the corresponding `ExtractedComposition` object that should not exist. 

This SQL command will destroy all the unwanted `ExtractedComposition` records, leaving only the `ExtractedLMRec` objects:
```
DELETE FROM dashboard_extractedcomposition where rawchem_ptr_id IN (SELECT 
            rawchem_ptr_id
        FROM
            dashboard_extractedlmrec);
```

After running this on my machine, the LM editing worked as expected.

mwfrost:
I think the next troubleshooting step is to upload a new data group of Literature Monitoring records and confirm that they can be updated in the card editor. Do you have a backlog of LM data that can be used to test it? 

If a new csv upload works correctly, we can consider this duplication to be an artifact of the old code, and wipe out the duplicate composition records.

Sakshi-Handa:
Previously, we did not have a method for uploading LM data by script. We would just upload the document pdfs, and curators would manually extract the chemicals into the chemical cards via the editing tools on the /datadocument/ page.

I think the only LM data that can test it would be the new files Kristin would like to upload, the IFC extracted LM data. I dont know if that is ready yet. Could you just test it with 'fake' data for now?

mwfrost:
This helps clarify things, thanks! The old records were added as `ExtractedComposition` objects, but when migration `0196` converted them into `ExtractedLMRec` objects, the migration did not delete the `ExtractedComposition` records -- they remained associated with each of the `RawChem` base class records. I will create a data migration to clean them out, after which the editing will work.

PR #2025, opened 2021-07-08
2023 fix lm card edits

mwfrost:
Closes: #2023 

Two data groups containing Literature Monitoring documents were added by Factotum users who manually entered cards for the chemicals. Those cards were stored in the database as `ExtractedComposition` subclass records, since we had not yet created the `ExtractedLMRec` model that now holds the chemical-level details for LM documents. 

When we added the `ExtractedLMRec` model (which inherits from the `RawChem` base class), a SQL command in a data migration created a new `ExtractedLMRec` record for each `RawChem` record that had been added to the two data groups. That migration did _not,_ however, remove the old `ExtractedComposition` records that had been created for each card. When the card updating view saves edits made in the card UI, the update view code uses `get_subclass()` to save form changes to the appropriate `RawChem` subclass. In this case, it's supposed to be the `ExtractedLMRec` subclass, but because those `ExtractedComposition` records are still out there, it finds that subclass first and the changes don't get applied. 

The migration in this branch deletes any `ExtractedComposition` records that share a `rawchem_ptr_id` with an `ExtractedLMRec` record. After those rows are removed, the LM edits are saved as expected.


fang-yan-csra (approved):
Verified no more extractedcompositon linked to any LM type of documents after migrations. All tests passed.

PR #2027, opened 2021-07-12
2011: rebuild tags index when docker start up

### LINKED ISSUE #2011: Adding index command into Docker
Following `issues/humanexposure/factotum/1960`: it was determined that the issue was elasticsearch index hadn't loaded. Fix was simple, to run the mangement command that sets it up.

This is a ticket to add this command into Docker so that it's part of the workflow when the container spins up. While this could also be done in Jenkins, Docker would be preferred since this would involve DevOps less, and (potential) troubleshooting can be done from this side.

fang-yan-csra:
Closed #2011 

When factotum container starts up, run the command to rebuild dashboard_tags index.


mwfrost (approved):
Tested by building the `factotum` and `factotum_ws` containers as the production environment does. After `docker-compose up`, the factotum container's logs included the creation of the `dashboard_tags` index.

Issue #1986, opened 2021-06-15
Getting Started on Homepage


Sakshi-Handa:
As a Factotum admin
I would like to add a section "Getting Started" to the homepage 
So that I can communicate to users how to access data and navigate the site.

On the homepage, we want a place to explain what Factotum is, how to explore and search for data, how to submit comments, etc.  
The EcoTox homepage is a great example. https://cfpub.epa.gov/ecotox/

Can we make it editable in the admin panel. I feel like we will be adding to this frequently, as we build new features. Would I be able to format the text with headers, bullet points, add links, etc.? 

Definition of Done:
• Add new section for "Getting Started" to the Factotum homepage, https://factotum.epa.gov/. 
• Appears for logged-in and guest users
• Ability to add and edit text via the admin panel.

Sakshi-Handa:
In the Getting Started section, I would like the ability to add a pdf attachment. So we can add a "quick start" guide that users can download and follow along with.

PR #2032, opened 2021-07-19
1986 getting started

mwfrost:
Closes: #1986 

This branch uses the `ckeditor_uploader` library, so please run `pip install -r requirements.txt` before running the migration. The new library allows users to upload images and files from inside the rich text editor. 

Once the new editor is installed and the new `section` field has been added to the `News` model, create a new News record in the admin panel and assign it the "gettingstarted" section. Test out whatever formatting you think might be applied, and upload a pdf. The pdf can be included as a preview (not a great option, since it takes so much space) or can be linked to after upload.


fang-yan-csra (approved):
Verified admin can edit getting started page which will display on homepage.

Issue #2007, opened 2021-06-29
Download Harmonized Media


Sakshi-Handa:
As a Factotum admin,
I would like to add a download for harmonized media
So that users can have access to this information. 

Harmonized media are located here:  https://factotum.epa.gov/admin/dashboard/harmonizedmedium/
Descriptions will be added in #2006

Definition of done:
Add a "Download Harmonized Media" button to the "download metadata" section of the /get_data/ page, allowing users to download a .csv file of Harmonized Media and their definitions

PR #2033, opened 2021-07-19
2007 download harmonized medium

BryanMG957:
Closes: #2007


mwfrost (approved):
The metadata download works as directed, and the test confirms the presence of a medium loaded in migration 0196

PR #2034, opened 2021-07-22
2008 harmonized media index page

BryanMG957:


fang-yan-csra (approved):
Harmonized media index page works well.

PR #2037, opened 2021-07-26
2009: schedule a task to generate the bulk composition data

### LINKED ISSUE #2009: Fix Composition Bulk Download
The Bulk download for Composition data created in #1964 is not working.
The download takes 30 seconds to run. The resulting csv file is empty, except for column headers. 

We can try another solution, such as switching to a nightly updating zipped file.

Definition of Done:
The bulk composition download contains useable data.

fang-yan-csra:
Closes #2009 

1. Added a nightly job to generate the `composition_chemicals.zip` file to download.
2. The cron job schedule is set by `GENERATE_BULK_DOWNLOAD_SCHEDULE` environment variable, and defaults to 4am `0 4 * * *`
3. To run celery beat locally, open another terminal and run `celery beat -A factotum`


mwfrost (approved):
This works great - I tried downloading the file before running the script (with prod data), and it reported that the data was not yet available. Then I forced the file creation by running the celery task, and the 68 MB zip file downloaded correctly. The csv was 1.3 GB.

The docker container for factotum will need to be rebuilt, and the new `GENERATE_BULK_DOWNLOAD_SCHEDULE` variable should be added to `.env` if you want anything other than the 4AM schedule.

Issue #2004, opened 2021-06-29
Search Bar Improvements


Sakshi-Handa:
As a Factotum admin
I would like to search by multiple keywords on the factotum main search bar
So that users can make more specific queries

Is there a way to improve the search so that users can search for specific phrases?
For example: searching "pet shampoo" returns these results. Not everything related to 'pet shampoo' is at the top of the page. There are also entries that just say 'pet' without ever mentioning 'shampoo'. 
https://factotum.epa.gov/search/product/?q=cGV0IHNoYW1wb28=

Definition of Done:
Ability to search for specific phrases in the Factotum main search bar.
prioritize phrases found in product name, document title, etc. over the description.

PR #2057, opened 2021-08-13
2004 search query

mwfrost:
Closes: #2004 

This branch changes the MultiMatch search functions in `factotum/elastic/search.py` to use `type="phrase"` when the search term is quoted. 

I have also added a lot of code to `test_search.py`. 
The search index is totally separate from the database, so we have never tried to index any of the ephemeral data created or updated during a test. Now my new test creates twelve documents with factories, uses the Logstash SQL query to select them with their related info, and then converts the results to a JSON request that populates the index. The `tearDown()` method removes the new records. 

If you want to add over a hundred thousand records to your elastic index for testing, run this in a shell with the prod database selected in your `.env` file:
```{python}
    from django.conf import settings
    import base64
    import requests
    esurl = settings.ELASTICSEARCH["default"]["HOSTS"][0]
    index = settings.ELASTICSEARCH["default"]["INDEX"]
    (es_username, es_password) = settings.ELASTICSEARCH["default"][
        "HTTP_AUTH"
    ]
    auth_header = {
        "Authorization": "Basic "
        + base64.b64encode(
            f"{es_username}:{es_password}".encode("utf-8")
        ).decode("utf-8")
    }

    # PUC = 259 (Pet shampoo) or PUC = 210 (shampoo)
    docs_json = fetch_dashboard_logstash(where="WHERE puc_id IN (259, 210)")
    for doc_dict in docs_json:
        # add the JSON to the index
        response = requests.post(
            f"http://{esurl}/dashboard/_doc/",
            json=doc_dict,
            headers=auth_header,
        )
```
I have included that snippet as a method in `elastic_factories.py`. 

**WARNING:** it will take a while. Append ` LIMIT 50000` to the `WHERE` clause if you want to load an arbitrary subset.

`factotum/elastic/search.py` uses the `PHRASE_SLOP` constant to decide how many other terms can interrupt the terms in the search phrase. Here are some examples of how the variable influences the results. We can fine-tune them in the future.

`PHRASE_SLOP = 0`
image

`PHRASE_SLOP = 1`
image

`PHRASE_SLOP = 2`
image


zachmunzenrider:
I had to add `from dashboard.tests.elastic_factories import fetch_dashboard_logstash` to get this to run. Seems to be trucking away now

PR #2058, opened 2021-08-17
add hover help text datadocument page

BryanMG957:


zachmunzenrider (changes requested):
Looks great.  I do think **Statistical Values** needs hover text also, even though it doesn't match up perfectly with the text provided. Perhaps, "Measurement statistics, with an indicator that value was reported in study or calculated by data curators"?

PR #2065, opened 2021-08-19
2061: Fix the Skip and Exit buttons' conditional behavior

mwfrost:
I didn't get the conditional behavior of the "Skip" and "Exit" buttons on the data document QA page right in a previous ticket. 

If a document belongs to a Composition data group and was manually extracted, the "Exit" button should return the browser to `/qa/manualcomposition/x/` where `x` is the data group ID. 

If it was extracted with a script, the Exit button should go to `/qa/extractionscript/y/` where `y` is the extraction script's ID. 

If it's a Chemical Presence document, the target should be `/qa/chemicalpresencegroup/z/` where `z` is the data group ID. 

The business logic has been moved from template conditions into the `qa.py` view, and there's a new test to catch a few of these conditions.


zachmunzenrider (approved):
Really satisfying refactoring. Good tests, clear code.

Issue #2005, opened 2021-06-29
Error Searching for Symbols


Sakshi-Handa:
When using the Factotum main search bar, certain symbols such as forward slash /, parentheses ( ), do not work in the search and return a 500 error.
For example, searching for this product: https://factotum.epa.gov/product/669172/
Typing in 'normal/ oily' in the search returns a 500 server error. 

Also not able to search for chemical names with ( ). 
For example '(-)-beta-Pinene' is not searchable.

And some symbols are ignored in the search:  $%&
For example, searching for this product: https://factotum.epa.gov/product/98591/
Typing in 'bath $10' in the search, and it ignores the $
https://factotum.epa.gov/search/product/?q=YmF0aCAkMTA=

mwfrost:
The special characters need to be escaped in `search.py`. Entering `normal\\/ oily` into the search box returns the expected results.

PR #2071, opened 2021-08-25
2005 escape search characters

mwfrost:
Closes: #2005 

Special characters in query terms are now escaped before they are sent to the elasticsearch server. See the original ticket for examples.

I have refactored the functional tests into a new class that passes a factory-built request to the `search_model` view rather than using the test client. This helps avoid just reproducing a bunch of string-encoding logic inside the test.


zachmunzenrider (approved):
Works great, tests all pass!

PR #2094, opened 2021-09-24
Changed bulk assingment lower limit

BryanMG957:
Changed to 5 after discussing with Sakshi, might lower again to 3 later.


mwfrost (approved):
Functionality and tests have been updated, working properly.

Issue #2172, opened 2022-01-26
BA Assigned products not appearing on Products Table


Sakshi-Handa:
There is an issue viewing Bulk Assign Raw Category (BA) products on the Products Table tab on the /chemical/, /puc/#, and /functional_use_category/# pages.
I believe this is a new issue, appearing since #2088 was completed. Previous BA assignments were working correctly. 

For example. 
on the products tab of the chemical page. 
https://ccte-factotum.epa.gov/chemical/DTXSID2021319/
the product 'zep formula 300'  appears as Automatic assignment to
"Home maintenance - degreaser".
But when you click on the product detail, the UberPUC assignment is BA Assignment to "Cleaning and safety - cleaning products - industrial degreaser"
The 'zep formula 300' product doesn't show up on the PUC page for Industrial Degreaser. Which we would want, since that is the UberPUC assignment.
It looks like the PUC appears on the product page as BA assigned to industrial degreaser, but it is treated on the product tables as if there is only the automated assignment.

Definition of Done:
BA assigned products should appear as a higher classification method than AU assignment on Product Tables visible on the /chemical/, /puc/#, and /functional_use_category/# pages.

mwfrost:
Some PUC assignments have been added without the system catching them and updating the `is_uber_puc` attribute for the product ID. 

The solution has two parts: 
1. Identify the place where new `ProductToPUC` objects were created without the `update_uber_puc()` method being run on the product.
2. Establish a bulk routine, maybe in celery, that will guarantee fresh uberpucs every day.

mwfrost:
Query to identify all the misassigned ones:
```
SELECT 
	ptp.product_id AS product_id,
	ptp.puc_id AS puc_id,
	ptp.classification_method_id AS classification_method_id,
    is_uber_puc,
	cm.rank AS rank,
	 (
			SELECT MIN(cmsub.rank)
			FROM 
				dashboard_producttopuc ptpsub
			LEFT JOIN dashboard_producttopucclassificationmethod cmsub ON cmsub.id = ptpsub.classification_method_id
            WHERE ptpsub.product_id = ptp.product_id
            GROUP BY ptpsub.product_id
    ) as min_rank,
    cm.rank =  (
			SELECT MIN(cmsub.rank)
			FROM 
				dashboard_producttopuc ptpsub
			LEFT JOIN dashboard_producttopucclassificationmethod cmsub ON cmsub.id = ptpsub.classification_method_id
[...]

PR #2182, opened 2022-02-07
2172 uberpuc cleanup

mwfrost:
Closes: #2172 

This turned into a larger-than-expected ticket. 

The root of the problem is that there are two ways of identifying which PUC assignment is the uberpuc: using python logic in real time, or using the hard-coded `is_uber_puc` attribute on the `ProducToPUC` model. The attribute was added to allow high-performance database queries, and it needs to be kept in sync every time the PUCs assigned to a product are changed. We use "signals" in Django to react to every change in the `ProductToPUC` model, but it turns out that the `bulk_create()` and `bulk_update()` methods used on the raw category assignment page bypass the signals. The endpoint was creating new `ProductToPUC` records and updating old ones, but the uberpuc refresh was not happening. 

In the original bug report, the user noticed that while the "BA" assignment was shown on the Product detail page, other displays were still showing the "AU" assignment as the uberpuc. This discrepancy occurred because the Product detail page derives the uberpuc in real time, while nearly every other screen that shows a product-to-PUC assignment uses the attribute for performance reasons. 

To see the PUC assignments that were never updated to reflect the uberpuc change, and which raised the problem cited in the ticket, run this query:
```
SELECT 
	ptp.product_id AS product_id,
	ptp.puc_id AS puc_id,
	ptp.classification_method_id AS classification_method_id,
    is_uber_puc,
	cm.rank AS rank,
	 (
			SELECT MIN(cmsub.rank)
			FROM 
				dashboard_producttopuc ptpsub
			LEFT JOIN dashboard_producttopucclassificationmethod cmsub ON cmsub.id = ptpsub.classification_method_id
            WHERE ptpsub.product_id = ptp.product_id
            GROUP BY ptpsub.product_id
    ) as min_rank,
    cm.rank =  (
			SELECT MIN(cmsub.rank)
			FROM 
				dashboard_producttopuc ptpsub
			LEFT JOIN dashboard_producttopucclassificationmethod cmsub ON cmsub.id = ptpsub.classification_method_id
            WHERE ptpsub.product_id = ptp.product_id
            GROUP BY ptpsub.product_id
    ) as should_be_uberpuc
	FROM
		dashboard_producttopuc ptp
	LEFT JOIN dashboard_producttopucclassificationmethod cm ON cm.id = ptp.classification_method_id
	WHERE is_uber_puc <> (cm.rank = (
			SELECT MIN(cmsub.rank)
			FROM 
				dashboard_producttopuc ptpsub
			LEFT JOIN dashboard_producttopucclassificationmethod cmsub ON cmsub.id = ptpsub.classification_method_id
            WHERE ptpsub.product_id = ptp.product_id
            GROUP BY ptpsub.product_id
    ))
    ORDER BY product_id, rank
    ;
```
The results will look like this:
image

In this table, you can see that many PUCs have been assigned to Products with both the "AU" assignment method and the "BA" method, but their `is_uber_puc` attribute is still assigned to the "AU" record. 

To test this branch, pick one of the mistakenly assigned products and look at its detail page:
`http://127.0.0.1:8000/product/702864/`
The PUC shown there is the actual uberpuc, even though the `is_uber_puc` attribute disagrees. 
Click the link to the product's Document Detail page to see the raw category and data group:
image

Find the corresponding row for the data group + raw category combination on the batch assignment page:
image

After selecting that row, click the "Connect PUCs" button to assign some new PUC to the records that you don't expect to already be matched up to some chemical in the product. 
image

The final step is to make sure that the new PUC assignment is reflected not only on the Product detail page, but on some page that uses the `is_uber_puc` attribute. One way to do this is to return to the data document detail page and choose one of the less common chemicals that appears in the ingredients. In this case, Decyl D-Glucopyranoside: `http://127.0.0.1:8000/chemical/DTXSID30893008/`

So far, so good:
image

Also check the "Products" table to confirm that the product's PUC has been updated:
image

This is the correct behavior.

*Fixing old issues*

Fixing the real-time uberpuc assignment will only prevent the problem going forward. The old mistakenly assigned records need to be fixed.

Once you have reproduced the error and seen in detail what went wrong before, you can run the new celery script in a shell to fix all the old records:
```
from dashboard.tasks import uberpuc_cleanup
uberpuc_cleanup.apply()
```
After that task finishes, try the SQL query again and there should be zero records returned.


dashboard/forms/rawcategory_to_puc.py
```diff
@@ -13,6 +14,11 @@ class RawCategoryToPUCForm(forms.Form):
         .distinct(),
         to_field_name="raw_category",
     )
+    created_by_id = forms.IntegerField(widget=forms.HiddenInput())
```
mwfrost, line 17: This addresses the comment in the ticket about how PUCs were being assigned to products without any of the metadata being updated.

dashboard/forms/rawcategory_to_puc.py
```diff
@@ -59,20 +65,31 @@ def save(self):
...
-                        product=product, classification_method_id="BA", puc=puc
-                    )
+                product_to_puc.updated_by_id = self.cleaned_data.get("created_by_id")
+                product_to_puc.puc_assigned_usr_id = self.cleaned_data.get(
+                    "puc_assigned_usr_id"
                 )
+                product_to_puc.updated_at = self.cleaned_data.get("created_at")
+                product_to_puc.save()
 
-        ProductToPUC.objects.bulk_create(product_to_puc_requires_create)
```
mwfrost, line 74: We were using the bulk_create() and bulk_update() methods to speed up the assignments, but that prevented the uberpuc update signal from launching. So I have fallen back on save() and update().

dashboard/models/product_to_puc.py
```diff
@@ -72,6 +72,9 @@ def update_uber_puc(self):
     class Meta:
         unique_together = ("product", "puc", "classification_method")
 
+    def shorthand(self):
```
mwfrost, line 75: I added this method because while I was testing I wanted a quick output format that described the ProductToPUC record.

dashboard/tests/functional/test_rawcategory_to_puc_form.py
```diff
@@ -1,4 +1,5 @@
 from uuid import uuid1
```
mwfrost, line 1: This definitely was too big to be a unit test, so I moved it to `dashboard/tests/functional`

dashboard/views/rawcategory_to_puc.py
```diff
@@ -110,7 +111,16 @@ def post(self, request):
             return HttpResponseRedirect(reverse("rawcategory_to_puc"))
         for datagroup_rawcategory_group in datagroup_rawcategory_groups:
             datagroup_rawcategory_group.update({"puc": request.POST.get("puc")})
+            datagroup_rawcategory_group.update({"created_by_id": request.user.id})
```
mwfrost, line 114: Updating the metadata

dashboard/models/product.py
```diff
@@ -1,6 +1,7 @@
 from taggit.managers import TaggableManager
 
 from django.apps import apps
+from django.db import connection
```
hausman-gdit, line 4: leftover?

mwfrost: Looks like it. Also `from .data_source import DataSource` and `from django.db.models import Max` could go, according to my VSCode linter.

hausman-gdit (approved):
Found bad records, ran fix, no bad records, tests pass. Looks good, thanks!

PR #2187, opened 2022-02-16
2179 tld env updates

kbfepa:
- Minor cleanup/adjustments to hostnames
- On dev and staging servers, I have adjusted `MEDIA_URL`, `MEDIA_ROOT` to the right values, so once merged they will point to their own respective media URLs
- I've also asked infrastructure team to copy the `/media` contents from production server to staging/dev


mwfrost (approved):
Changes all make sense.

PR #2188, opened 2022-02-17
2174 Correct summary stat

mwfrost:
Closes: 2174

This has already been demonstrated in a review call.


dashboard/models/harmonized_medium.py
```diff
@@ -25,12 +26,19 @@ def get_absolute_url(self):
...
+                .filter(Q(unionextractedlmhhrec__harmonized_medium=self))
+                .values_list("dsstox__id", flat=True)
+            )
+            mmdb_chems = (
+                MediaSampleSummary.objects.filter(dsstox__isnull=False)
+                .filter(Q(harmonized_medium=self))
+                .values_list("dsstox__id", flat=True)
+            )
+            vals = list(factotum_chems) + list(mmdb_chems)
+            return DSSToxLookup.objects.filter(id__in=vals).count()
```
hausman-gdit, line 40: Does this sort of act like a second catch to what could be missed in the `dsstox_isnull` filter?

mwfrost: The `DSSToxLookup` records are the foreign keys to two different tables, and this ticket is about showing the intersection of records in those two tables. The `factotum_chems` list stores all the SIDs that appear in the `RawChem.dsstox` column, while `mmdb_chems` stores all the SIDs that appear in `MediaSampleSummary.dsstox`. There will be a lot of duplicates. The `vals` list represents the concatenation of the two lists from the two sources, and then the final query in line 40 uses that list to select distinct rows from `DSSToxLookup`.

dashboard/tests/functional/test_ajax.py
```diff
@@ -395,6 +395,11 @@ def test_chemicals_by_harmonized_medium(self):
         data = json.loads(response.content)
         self.assertEquals(data["recordsTotal"], 6)
 
+        # The union of MMDB and RawChem records should also be reflected in the
+        # chemical_count property of the harmonized medium model
+        hm = HarmonizedMedium.objects.get(pk=4)
+        self.assertEquals(hm.chemical_count, 6)
```
hausman-gdit, line 401: Works - test fails as expected when adding another level of filtering onto the MMDB count, passes when reversed

dashboard/models/harmonized_medium.py
mwfrost: It might be the case that `.filter(dsstox__isnull=False)` is redundant, but I wanted to reduce the number of `RawChem` and `MediaSampleSummary` records that were being compared in the subsequent filters. I _think_ this helps optimize the final SQL query, maybe not.

hausman-gdit (approved):

dashboard/models/harmonized_medium.py
hausman-gdit: sets?

mwfrost: 🤔

Issue #2098, opened 2021-10-06
Search Tab for Harmonized Media


Sakshi-Handa:
As a Factotum user
I would like to use the main search bar to search for harmonized media categories
So that I can view search results related to my harmonized media of interest

Definition of Done:
	• Applies to the Factotum main search, https://ccte-factotum.epa.gov/search/product/?q=aW5kb29yIGFpcg==
	• Create a new tab for 'Media' along the top of the page, similar to the CP Tag search tab. 
	• Search results in 'Media' tab displays Harmonized Media name and definition, and links to the Harmonized Media detail page.
	• Searching by a keyword, i.e. 'air' should look through harmonized media categories and their descriptions and return all relevant results.

mwfrost:
@Sakshi-Handa 
Adding two wide tabs to the list of search categories at the top has used up the remaining space, and now the `<li class="nav-item">...</li>` elements have wrapped around into a second row. Please add a new ticket if you want us to redesign the navigation tabs to account for how popular that real estate is.

mwfrost:
![image](https://user-images.githubusercontent.com/201835/156251990-1e7f3769-5713-4379-865a-5a31b1886d28.png)

Issue #2099, opened 2021-10-06
Search Tab for Functional Use


Sakshi-Handa:
As a Factotum user
I would like to use the main search bar to search for functional use categories
So that I can view search results related to functional uses of interest

Definition of Done:
	• Applies to the Factotum main search, https://ccte-factotum.epa.gov/search/product/?q=YWJyYXNpdmU=
	• Create a new tab for 'Functional Uses' along the top of the page.
	• Search results in 'Functional Uses' tab displays the name and definition, and links to the FU detail page.
	• Searching by a keyword, i.e. 'abrasive' should look through Functional Use Categories and their descriptions and return all relevant results. 
	• Search results should be flexible.

PR #2198, opened 2022-03-01
2098 Harmonized Media search

mwfrost:
Closes: #2098 
Blocked by: #2099 

As with #2099 , test this by running `python manage.py search_index --rebuild` and then searching for a term from the harmonized media names. Use the admin panel to add some text to a description, rebuild the index again, and search for a newly-added keyword.


hausman-gdit (approved):
added text and it got picked up, tests pass, works as expected

PR #2206, opened 2022-03-07
2205 disable autocomplete

kbfepa:
To check: spin up the application, and on `/login`, login with creds. Navigate back to `/login`, and when clicking on input box for username/password, there shouldn't be suggested autocomplete popping up. 
(Checked with Chrome.)


mwfrost (approved):
As long as the browser is set to respect the `autocomplete` tag in the form (as on EPA's network), this will work.

Issue #2200, opened 2022-03-02
MMDB Summary Table Appearance


Sakshi-Handa:
As a Factotum admin, 
I would like to improve the appearance of the MMDB summary tables, 
In order to better display the data for users. 

Ex. https://ccte-factotum.epa.gov/harmonized_medium/29/
https://ccte-factotum.epa.gov/chemical/DTXSID9032119/
On the MMDB summary table, longer entries for the 'data source' field cause the rows to expand in order to fit the data. It makes the table difficult to read in some cases, and results in a lot of 'empty space'. 

![image.png](https://images.zenhubusercontent.com/5f318c2978fa23cbc1a57b71/b6a3c5ef-887f-4e0a-80bc-c6ed5ea16201)

Definition of Done:
	• Applies to the MMDB summary tables on the /harmonized_medium/ and /chemical/ pages
	• Adjust the table rows/columns to better fit the large 'data source' entries and improve table appearance

mwfrost:
We discussed truncating the Source Name field and presenting the full string in a tooltip. Is that an acceptable solution? The other eight columns cannot be condensed any further.

PR #2217, opened 2022-03-14
2200 Truncate DataTables field

mwfrost:
Closes: #2200 

This branch uses the "ellipsis" DataTables plugin to truncate the SourceName column to eight characters.


hausman-gdit (approved):
Fixes issue, useful addition, mild UI inconsistency

PR #2225, opened 2022-03-28
2222 puc detail hover format change and addition

leogi23:


mwfrost (approved):
The `popper.js` tag works perfectly, and the new help text matches what was specified in the ticket.

PR #2229, opened 2022-03-29
1774 datagroup performance

### LINKED ISSUE #1774: Data Groups Page Performance
As a data curator
I would like to improve the performance of the datagroups page
So that I can easily search the list of Factotum datagroups 

https://factotum.epa.gov/datagroups/
The data groups page was recently updated to AG Grid. 
The page is slow to load, and returns an 'page unresponsive' message when attempting to interact with the table. 
It eventually loads after about a minute, and the table functionality works well.  

Definition of done:
Improved performance and responsiveness of /datagroups/ page

hausman-gdit:
Closes #1774.

Column sizing and row height are set for the entire dataset while being paginated. I believe there was a 'cartesian product' of sorts where each new datapoint triggered a cascade of resizing.

Removed the `onColumnResized` listener function and did a pseudo loop to set appropriate row height and column sizing. 

I'm not sure if this is testable? I used production data to force the performance issues.


mwfrost (approved):
This rendering change solves the main performance issue, and I agree that a test is not advisable here. 
The query itself takes around 3 seconds with production data, which seems like something we could optimize later.

PR #2232, opened 2022-04-01
2224 point to local dockerfile

### LINKED ISSUE #2224: Point to local Dockerfile for builds for dockerized development
Currently, `docker/docker-compose.overrride.yaml` has services that point to remote repo URLs. However, because it points to a remote repo's copy of the Dockerfile (on a specified branch), it doesn't allow local code changes to be picked up without first being pushed.

Acceptance criteria:
- instead of github URL, should point to Dockerfile in the repository, locally
- check that this takes advantage of caching mechanism during build process

kbfepa:
Closes #2224 .

To test: make any noticeable changes to any of the services, `docker-compose rebuild <service>`, `docker-compose up -d`, and the said change should be observable. Previously, this wouldn't have been the case unless `docker-compose rebuild <service> --no-cache`, after having pushed the relevant branch, so that docker can pick up the source code from git.


hausman-gdit:
This is sort of two fold then; one to fix your particular issues with github/connectivity and then prepping for however the pipeline will change post migration?

kbfepa:
Previously, all of the following had to happen to observe local code change, every single time... (this includes the most simple things, like capitalizing a letter on an HTML file).
(1) Make code change
(2) Push code change to remote Github on `<some_feature_branch>`
(3) Adjust `.env` such that `FACTOTUM_BRANCH=<some_feature_branch>`
(4) `docker-compose build --no-cache <service_name>` (which builds from Dockerfile/code on remote repository)
(5) `docker-compose up ...`
Which takes about 10-15 minutes total, because `--no-cache` takes a few minutes.

With this branch, this is reduced to:
(1) Make code change
(2) `docker-compose build <service>`
(3) `docker-compose up ...`
Which takes about 10-30 seconds.

hausman-gdit (approved):
Approved!

PR #2246, opened 2022-04-20
2236 qa verified checkbox

### LINKED ISSUE #2236: QA Verified Checkbox on QA Summary
As a Factotum admin
I would like the ability to verify if the QA completed for a certain QA Group with a checkbox 
So that I know when QA is of sufficient quality.

Definition of Done:

- Applies to the QA Summary Page for all QA Workflows:

	• Composition Extraction Script (https://ccte-factotum.epa.gov/qa/extractionscript/52/summary)
	• Composition Manual (https://ccte-factotum.epa.gov/qa/manualcomposition/634/summary)
	• Functional Use Extraction Script (https://ccte-factotum.epa.gov/qa/extractionscript/322/summary)
	• Chemical Presence (https://ccte-factotum.epa.gov/qa/chemicalpresencegroup/801/summary/)
	• Composition Cleaning Script (https://ccte-factotum.epa.gov/qa/compositioncleaning/316/summary)

- Add a checkbox on QA Summary page for "QA Verified", above the QA Summary Note text box.

leogi22:
Closes #2236 
Addition of QA Verified checkboxes in several data group/scripts operations


static/js/dashboard/qa_summary_qa_check.js
```diff
@@ -0,0 +1,15 @@
+$(document).ready(function() {
```
hausman-gdit, line 1: This occasionally breaks, I think we need to push this back down into django's url stuff, something like what the Notes field does

templates/qa/chemical_presence_summary.html
```diff
@@ -18,6 +18,8 @@ <h1>QA Summary: {{ datagroup.name }}</h1>
         <dd class="col-9" id="qa_notes">{{ datagroup.qa_note_count }}</dd>
         <dt class="col-3">Total Remaining Documents</dt>
         <dd class="col-9" id="qa_incomplete_count">{{ datagroup.qa_incomplete_count }}</dd>
+        <dt class="col-3">QA Verified</dt>
```
hausman-gdit, line 21: We can access `datagroup.qa_verified` directly and then use the unique id of the element to trigger/key in on,

something like:`<input type="checkbox" id="qa_approved-{{ datagroup.pk }}" value={{datagroup.qa_verified}} checked>`

hausman-gdit, line 21: We can access the `datagroup.qa_verified` directly instead of packaging it separately as checked - its True/False

The checkbox `onChange` can trigger your post and leave the rendering to django and out of the url pattern, something like this maybe:

`<input type="checkbox" id="qa_approved-{{ datagroup.pk }}" onchange="MyCustomFUnction()"  value="{{ datagroup.qa_verified }}" {% if datagroup.qa_verified %} checked{% endif %}>

hausman-gdit (approved):
Approved! 
I think from a 'djangoization' standpoint a nick-pick would be that the checkbox status setting maybe should be like an inherited property of the object? iirc then the `datagroup.qa_...` and `scripts.qa_...` may not be necessary?

PR #2254, opened 2022-04-25
2234 dspl product guest

### LINKED ISSUE #2234: Display of product attributes for guest user
As a logged out user,
I only want to view attribute information for a product if there is curated attribute information present, 
So that I am not confused by allowed attribute options. 

Ex. https://ccte-factotum.epa.gov/product/557581/
For this air freshener product, no allowed attributes are selected from the available options (liquid, aerosol, get, etc.) We do not want to show guest users all of the allowed attributes options, when this data hasn't been curated or assigned an attribute.  

Definition of Done:
- Applies to the /product/# page for a guest/logged out user
- Only show the assigned attribute if there is a selected attribute present for a given product. Do not show all allowed attribute options. 
- Only show the header for Assumed and Allowed Attributes if there are attributes associated with the product. If there are no attribute tags selected under 'Assumed Attributes' or 'Product Attributes', hide these headers from the guest user.

leogi22:
closes #2234


hausman-gdit:
This `(edit)` redirects to the django admin page (to edit _all_ product attributes) when you're not logged in. I think it should be fixed or just hidden as well.
image

hausman-gdit (approved):
Approved -  @hausman-gdit to be keelhauled and forever immortalized by failure.

PR #2255, opened 2022-04-28
2220: puc removal message

### LINKED ISSUE #2220: Messaging upon AU PUC Removal
Relates to recently completed #2173, for removing AU PUC predictions via csv upload. 
The upload works well, when we upload a csv with a blank puc_id, Factotum removes the AU PUC assignment from the related product_id with incorrect PUC assigned. 

But upon upload, the tool returns a message to the curators summarizing changes made. The messaging is not showing that pucs were removed, returning 'undefined PUC assignments added, undefined updated, undefined deleted.'

Definition of done:
- applies to AU PUC upload tool, /upload_predicted_pucs/
- after uploading csv file, the confirmation message should alert users of AU PUC assignments removed.

kbfepa:
Closes #2220 .
Previously, on `/upload_predicted_pucs`, an upload that only removes PUC assignment errored out because the variable `created` was being referenced without being defined (that variable only got defined when there was a creation of PUC assignment.)


hausman-gdit (approved):
Approved!

PR #2276, opened 2022-05-11
2181 dynamic reindex schedule

### LINKED ISSUE #2181: set logstash pipeline schedule to 15 minutes after deployment
TODO: still needs to be flushed out - ask Matt about it.

Instead of setting to 1:30AM, could there be a bit of string manipulation to dynamically set the `REINDEX_SCHEDULE` ? to `current_time + 15 min`?

=> There are many times when we want the index set immediately, but having them run more frequently than once a day would be wasteful.

kbfepa:
Closes #2181 .
This PR doesn't do anything on its own - until its twin PR in the other repository is completed, it won't take effect, since build args have to be declared in the Dockerfile. It offloads the dynamic computation to the host machine (via Jenkins), so any users will have to call the docker-compose build command with similar build arg.


hausman-gdit (approved):
Approved

PR #2285, opened 2022-05-23
2282-defaulting to document instead of product when searching

### LINKED ISSUE #2282: Change Default search tab to 'Documents'
As a factotum admin
I would like our users to see the Documents tab first upon searching
Since it is the search that provides the most general results for users. 

Currently, the Products tab is the default. When a user searches a term, the page automatically directs them to the 'Products' tab. The 'Documents' tab returns the most general results, so it would display everything we have in Factotum related to a user's query. 

Definition of Done:
	• Applies to the Factotum main search 
	• Change the default search tab from 'Products' to 'Documents'. Users should be initially directed to the "Documents" tab.
	• Change the page layout so that the 'Documents' tab appears on the left side as the first tab.

leogi22:
Closes #2282 

### Changes:
+switches order product<-->Document in search result page


hausman-gdit (approved):
Approved

PR #2286, opened 2022-05-23
closing tagset selection window once an item has been selected

### LINKED ISSUE #2278: Automatically Collapse Tag List on Selection
Applies to recently completed CP tag filtering ticket #1968.
I noticed that we have a lot of cp tag sets in the dropdown list. And users can only select one at a time. So I would like to make it easier for users to see the data, and make the application more 'responsive' by taking the user straight to the table of their filtered results. 

Definition of Done:
Applies to the "Associated List Presence Keyword Sets" on the /list_presence_tag/#/ page.
When a user selects a filter from the dropdown list, the list selection will automatically collapse.

leogi22:
Closes #2278 

### Changes:
+ added click to close tagset window upon selection (lp_tag_tables.js)


hausman-gdit (approved):
Approved

PR #2287, opened 2022-05-25
2284 mod puc prod search

### LINKED ISSUE #2284: Modify the 'PUC'  and 'Products' Search
As a factotum admin
I would like the 'PUC' and 'Products' search to only return search results relevant to PUCs and their definitions
So that users are not confused by complicated search results.

Currently, the PUC search tab returns chemicals that are found within products that have been curated to a specific PUC. I would like the 'PUC' search to work similarly to the Functional use and Media tabs, where we are only searching for the term name and definition. 

Definition of Done:
	• Applies to 'PUC' tab on Factotum main search  https://ccte-factotum.epa.gov/search/puc/?q=MSw0LURpb3hhbmU=
• Applies to 'Products' tab on Factotum main search https://ccte-factotum.epa.gov/search/product/?q=MSw0LURpb3hhbmU=

	• The PUC search should only search PUC category names and their definitions.  It should NOT search PUCs for underlying chemical information (i.e. chemical name, CAS, DTXSID).

	•  The product search should only search product names, product details and PUC information. It should NOT search products for underlying chemical information (i.e. chemical name, CAS, DTXSID).

hausman-gdit:
Closes #2284

This search function change may have unexpected consequences vis-a-vis parent/child groupings results 'deeper' in the application and out of scope from testing (e2e)

### Changes:
- Elasticsearch fields
- modified tests, removed one redundant test (similar redundancy noticed in passing elsewhere)


dashboard/tests/functional/test_search.py
```diff
@@ -106,45 +104,17 @@ def test_pagination(self):
         )
         self.assertEqual(element_count, 40)
 
-    def test_number_returned(self):
```
hausman-gdit, line 109: This test was redundant.

dashboard/tests/functional/test_search.py
```diff
@@ -296,43 +266,46 @@ def test_document_search_product_count(self):
             "TRENDstarter Hard Gel 5.1 fl oz has the incorrect number of associated products",
         )
 
-    def test_puc_search_product_count(self):
```
hausman-gdit, line 299: I'm not sure if this is supposed to test some sort of relationship resolution? I think the tag search functionality is quite 'deep'

leogi22 (commented):
is the ^4 describing that the search should follow the ORM 4 relationships deep?
It seems that only the main search uses Elastic Search code, i.e. only that very functionality that displays those results invokes that code, so if it's your understanding, there shouldn't be any worry regarding more implications

hausman-gdit:
> is the ^4 describing that the search should follow the ORM 4 relationships deep? It seems that only the main search uses Elastic Search code, i.e. only that very functionality that displays those results invokes that code, so if it's your understanding, there shouldn't be any worry regarding more implications

I found this describing it as a kind of 'search relevance' modifier, like this
> Use the boost operator ^ to make one term more relevant than another. For instance, if we want to find all documents about foxes, but we are especially interested in quick foxes:
quick^2 fox

leogi22 (commented):
looks like 7 tests in dashboard.tests.functional.test_search.TestSearch are now failing instead of just 1 before, due to the change in the search criteria.
I would recommend adjusting the expected numbers to match the new expectations.
If Sakshi needs it now, I can approve and merge, but the tests need to be revised while it's fresh in memory :)

hausman-gdit:
> looks like 7 tests in dashboard.tests.functional.test_search.TestSearch are now failing instead of just 1 before, due to the change in the search criteria. I would recommend adjusting the expected numbers to match the new expectations. If Sakshi needs it now, I can approve and merge, but the tests need to be revised while it's fresh in memory :)

7 failing tests?! I thought I'd already changed them all. I don't get any TestSearch errors hmmm

PR #2294, opened 2022-06-01
2260 add data release field to datagroups

### LINKED ISSUE #2260: Set Data Release flag via Admin
As a factotum admin
I would like to add a dropdown list to set data release flags for datagroups via the admin panel
So that I can manually set data release permissions for each datagroup. 

Ex. https://ccte-factotum.epa.gov/admin/dashboard/datagroup/908/change/

This ticket will be to add a new field to 'datagroups' datamodel. This new field should also appear on and be editable from the admin panel. 

Note: A future ticket will set accessibility of each datagroup on different environments in Factotum (private/curator, internal/guest, public application). The intention is to manually set data release flags based on curation level of each datagroup (i.e. we only want to release data that has been fully curated.) This field will be used as a criteria for data release to the public application and Comptox Dashboard. 

Definition of Done:
	• Add a field for 'Data Release' to the datagroup data model in the database
	• The field will have a controlled vocabulary, with 3 options: 'private', 'internal', and 'public'. The default for this field should be 'private'. 
	• Verify that the dropdown list for Data Release appears on the /datagroup/# page in the admin panel. An admin should be able to select one of the 3 options from the dropdown list, and save changes.
	• Add a new column for 'Data Release' on the /datagroups/ page in the application. This column will display the release flag for each data group. Curators can view but not edit the data release flags.

kbfepa:
Closes #2260 

### Changes:
- added "data_release" field to Datagroup model
- accompanying AG grid + Django view changes
- UI/html changes to navbar


leogi22 (approved):
removed unused variable in test

dashboard/tests/functional/test_datagroup.py
```diff
@@ -92,7 +92,6 @@ def test_data_release_in_table(self):
         response_html = html.fromstring(response.content.decode("utf-8"))
         tabledata = response_html.xpath('//*[@id="tabledata"]/text()')
 
-        datagroup_json = json.loads(tabledata[0])
```
kbfepa: nice catch, thanks!

PR #2297, opened 2022-06-06
2290 add heroid field

### LINKED ISSUE #2290: Add new field for HERO ID
As a factotum admin
I would like to add a new field for the HERO ID and allow curators to upload this info
So that I can store this metadata relevant to certain LM documents

Health & Environmental Research Online (HERO) is an internal EPA database with scientific articles of interest. (ex. https://hero.epa.gov/hero/index.cfm/reference/details/reference_id/1488252) This ID would be helpful to researchers trying to find data we have related to specific articles. 
The HERO ID primarily corresponds to certain literature monitoring (LM) type documents uploaded into Factotum. We're currently including the HERO ID in the document subtitle, but would want to create a separate field for it. 
We also want to add the Hero ID as a field associated with the document as part of the 'register records' csv when we upload documents to Factotum. I believe LM groups have a different upload template, that also includes the PubMed ID (PMID) field. 

Ex. LM document with HERO ID
https://ccte-factotum.epa.gov/datadocument/1625764/

Definition of Done:
	• Applies to LM type documents in Factotum
	• Add a new field to the datadocument model for 'Hero ID'. This should be a free text field (20 character limit).
	• Add a new column for the hero_id to the 'register records' csv upload. 
	• Add the Hero ID as an editable field on the data document edit form, https://ccte-factotum.epa.gov/datadocument/edit/1625764/

leogi22:
Closes #2290 

### Changes:
- Adds HEROId to datadocument model, 
- updates RegisterRecordTests
- updates DataDocumentCSVForm and DataDocumentForm Forms
**- REQUIRES to run migrations**


hausman-gdit (approved):
Approved

PR #2302, opened 2022-06-13
use redis for caching some views

### LINKED ISSUE #2251: Performance on Home Page
The Factotum homepage takes 5+ seconds to load. Wonder if this can be optimized to reduce load time. 

Definition of Done:
Please look into if there is any room to improve page performance of https://ccte-factotum.epa.gov/

Investigate 'caching' data for visualization pages.

kbfepa:
Closes #2251

### Changes:
- added a `resources.py` page where modified cache decorator is defined
- this Django cache decorator will be turned off if `DISABLE_CACHE_PAGE=1` is left in `.env`, which is the default for now

For devs, this default value should be there for testing reasons.


dashboard/resources.py
```diff
@@ -0,0 +1,22 @@
+"""File for defining resources/constants."""
+import os
+from django.views.decorators.cache import cache_page
+
+# cache resources for speed improvements in views
+def no_cache(*args, **kwargs):
+    # decorator that does nothing, taken from Terry Brown's work on GenRA
```
hausman-gdit, line 7: rm attribution

hausman-gdit (approved):
Approved

PR #2314, opened 2022-06-17
passing ids to url in a leaner way for tagset filtering

### LINKED ISSUE #2307: Adjust tagset filtering to use content_id_field
Currently the list presence tagset filter does not work on many tags, because the application uses string-ified form of Django objects (of all tag sets) to identify selected elements. This causes URL query string to be overextended, and errors out after exceeding maximum URL character length. It also leads to HTML elements with very long `id`s. This could be resolved by identifying each tagset filter with combination of (`tag_id`, `content_object_id`), which would result in simplifying the URL, HTML, and backend logic.

Acceptance criteria:
For each "code component" in {frontend/HTML, backend/Django, Ajax request/URL}, have each tagset filter row be identified with `tag_id` and `content_object_id` instead of raw string-ified form of Django objects.

leogi22:
Closes #2307 

### Changes:
- changes ajax, url, tags and tags_detail.html to pass ids to html then make lean ajax call


leogi22:
maybe what is there which works may be pushed/deployed so that it can be shown? until the other way is done?

kbfepa (approved):

PR #2317, opened 2022-06-21
2299 rm multi curated chem

### LINKED ISSUE #2299: Remove multiple entries via Curated Chemical Removal
As a Data Curator
I want to improve the curated chemical removal tool
So that I can remove multiple incorrect curated chemical linkages at once. 

We created the /curated_chemical _removal/ tool to remove the link between a raw chemical name/CAS and a curated chemical name/CAS/DTXSID everywhere it appears on Factotum. We use this page to remove incorrectly curated chemical assignments that we receive from ChemReg. 
In the example below, the extracted chemical of 'supplier trade secret' is incorrectly curated to the chemical 'Carbomer 980'. 
Currently, the page refreshes after a user selects the 'Remove Linkage' button. Then the user has to re-type the query and press 'enter' to reload the table. The tool remembers the filters you applied to the table, but it is inconvenient to reenter the search and wait for the table to load, and press the button for each incorrect entry. 

Ideas:
• Allow the user to stay on the same searched table and remove multiple entries at once. (don’t refresh the page after the button is pressed)
• Or add a checkbox along the side, so a user can select multiple entries and press a new 'remove all' button to remove all selected entries. 

Definition of Done:
• Applies to the /curated_chemical_removal/ page
• Allow user to remove multiple incorrect chemical linkages at once.

hausman-gdit:
Closes #2299

### Changes:
- Change from the 'universal' ajax view to a more appropriate subject specific view
- Add in checkbox to select 
- show/hide/POST


leogi22 (commented):
looks like when you select a few items on the first page and then navigate away to select other items in other pages and come back to check, the new first page didn't remember what had been selected. Maybe not in the ask idk but worth mentioning here if that was not intended

hausman-gdit:
> looks like when you select a few items on the first page and then navigate away to select other items in other pages and come back to check, the new first page didn't remember what had been selected. Maybe not in the ask idk but worth mentioning here if that was not intended

Crud bunnies, thanks!

PR #2318, opened 2022-06-23
2315 lm download rem aggregation and add columns

### LINKED ISSUE #2315: Bug in statistics column in LM download
In the recent ticket #2289, we added a download for the LM datadocument.
I was checking on stage, it looks like there is a bug in the 'stat's column. The download is combining all of the statistical values reported for the same 'chemical name' into one column. 
So in the example file, when two chemical records have the same 'raw chemical name', the 'stats' column combines the reported statistical values for each record and reports them twice on each row. 

i.e. https://ccte-factotum-stg.epa.gov/datadocument/1555997/#chem-card-5156465 
There are two records for 'decane' and the csv shows each record on its own row, but is displaying the statistics "Mean" and "LOQ" in the stats column for both records. Could  this be because it is combining records based only on the 'chemical name' column?

Here is the example csv file: [ex LM data download .csv](https://github.com/HumanExposure/factotum/files/8933733/ex.LM.data.download.csv) (I'm not sure the download is available anymore, since we rolled it back)

This is a screenshot of what it looks like on stg. 

Definition of Done:
• Applies to the LM document download
• Each chemical record should have its own row in the document, and should only display the statistic associated with that record. 
• I would like to change the display of this data, so that each section of the 'statistical values' has its own column. (i.e. 4 columns for the statistic name, value, unit and stat_type) 
• If there are multiple statistics associated with a single chemical record, display each statistic on its own row, duplicating the other chemical information.

leogi22:
Closes #2315 

### Changes:
- changed download_lm_monitoring_chemicals in data_document.py to display stats in columns and remove aggregation baed on raw_chem_name


hausman-gdit (approved):
There are a slew of failing tests but IIRC they're not all from this branch. Going to approve so it can go to dev for tomorrow

PR #2327, opened 2022-07-04
resetting curated data when raw data is changed

### LINKED ISSUE #2292: Auto-Remove Curated Composition Data
As a Factotum Curator
I would like the application to automatically remove curated weight fraction data when the raw composition data is edited
So that I can quickly remove incorrectly curated data and easily replace it

When curators are extracting or QA-ing the data on chemical records, we might need to edit incorrectly extracted raw composition data (raw minimum, central, maximum comp fields). If that data has already been curated to the cleaned weight fractions, then that cleaned data is also incorrect, So we want to also remove that as well. 
We have similar functionality already within Factotum, that removes the curated chemical name/CAS and DTXRID/SID when a user makes any edits to the raw chemical name/CAS.  (See ticket #914)
Ex. https://ccte-factotum.epa.gov/datadocument/1467036/

Definition of Done:
• Applies to the display of chemical card on composition data documents, and to the values within the composition data fields.
• When a curator edits a value in the raw chemical composition data fields (raw minimum, central, maximum comp and unit type), remove the curated composition data (cleaned lower, central, upper weight fraction).

leogi22:
Closes #2292 

### Changes:
- signal.py: signal added for subclass ExtractedComposition
- ExtractedComposition model: adding its own Field tracker
- Test_browser_edit: added ExtractedComposition  test
Please run test and report back as the results depend on chrome driver permissions


hausman-gdit (approved):
Approved

PR #2352, opened 2022-07-20
2330 adjust get_stats caching to be at function level using redis_lru

### LINKED ISSUE #2330: Bug: Displaying Incorrect Curator name
We were able to replicate the bug where the incorrect display name shows up for a logged in user.
Allison and I were logged in at the same time, while I was doing a demo on Factotum. 
We noticed that when you click on the home page, you appear as not logged in. 
Then when I clicked on the 'Statistics' tab, the display name in the corner changed to 'Allison'.
Allison was also on Factotum, and when she clicked on the 'Visualizations' tab, the display name changed to 'Sakshi'. 

I think this issue might be related to the "Performance on the Home Page" ticket. From what we can tell it only applies to these few pages that were cached. We weren't able to replicate this in any of the curation pages. 

Definition of Done:
Factotum displays the appropriate name corresponding to the logged in user.

kbfepa:
Closes #2330 

### Changes:
- moves the cache to python-method level for main page/stats
- adjusts `resources.py` so that it carries two types of cache: Django's native page-based cache, and redis_lru's method-based cache
- changes `.env` var name to turn of cache
- minor formatting changes via `black` linter and `isort` package


hausman-gdit:
@kbfepa What is `sort`?

hausman-gdit (approved):
Approved

hausman-gdit:
@kbfepa Hmm, can you shed some light on this merge conflict? I think the only part that may be 'missing' is this? 

`"puc_functional_uses_csv/<int:pk>/",
        views.download_puc_functional_uses,
        name="download_puc_functional_uses",
    ),`

hausman-gdit:
gitkraken_McAYL27I15

PR #2367, opened 2022-08-03
2356 adding cloning of chem cards in data documents

### LINKED ISSUE #2356: Ability to Duplicate Chemical Cards on Data Document
As a data curator
I would like the ability to duplicate data on existing chemical cards,
So that I can easily extract data for similar chemicals when manually extracting data. 

Ex. https://ccte-factotum.epa.gov/datadocument/276840/
In this document there are multiple components, with similar chemicals in each. So we manually extracted data for 'Water' 6 times. It would be easier if we could press a button to create a new chemical card with the same extracted data. Ideally, it would open the chemical edit form for the new record, and let you make any necessary edits and confirm the new record. 

Primarily interested in Composition data for now. In the future, we may want to expand to other data types. If it is easier, we can apply to all data types at once.

Definition of Done:
	• Applies to the /datadocument/ page for Composition type documents
	• Ability to select an existing chemical card, and copy the raw data within the card to create a new chemical record. We only want to copy the raw extracted data available on the chemical edit form, NOT the curated data (weight fractions, harmonized functional uses, etc).

leogi22:
Closes #2356 

### Changes:
- datadocument.py views: addition of new clone view
- co_cp_cards.html: updates with clone option


hausman-gdit (approved):
Approved

PR #2405, opened 2022-09-21
fixing resetting wf_ values when raw values changed

### LINKED ISSUE #2401: Bug: Document updating wf_analysis fields when manually editing chemical card
When manually editing a data document, the 3 wf_analysis fields update from none to 0.000 (lower_wf_analysis, central_wf_analysis, upper_wf_analysis), as shown below. 
I don’t think this was happening previously, it has shown up in production since the last deployment. 

I am getting an error related to the wf_analysis fields when I try to edit that chemical card again. (or even just press 'save' without making any edits to the chemical card). It doesn't allow the user to make edits/save the chemical card. 
Ex. https://ccte-factotum.epa.gov/datadocument/278870/

Definition of Done:
The wf_analysis fields should not be updating when a user manually edits the data. These fields are updated in a later curation step, where we harmonize the raw composition data to cleaned weight fractions.

leogi22:
Closes #2401 

### Changes:
- signal/py

Bug was that lower_wf_analysis, central_wf_analysis, upper_wf_analysis were reset to 0, which violated a data integrity check (central_wf_analysis can't be set when the other 2 are set). Resetting to None fixes that (it also makes more sense than 0 values)


hausman-gdit (approved):
Approved!

PR #2408, opened 2022-09-28
adding 'filename' into datadocument when adding new datadocument, to …

### LINKED ISSUE #2402: Bug: Manually uploaded data documents not registered in Factotum
When manually uploading new data documents, using the '+ new data document' button on the /datagroup/ page, 
The document is not saved in the same way as documents that are registered via csv. 

Ex. https://ccte-factotum.epa.gov/datagroup/335/
1. When you download the 'extracted text csv template', 
The data_document_filename field is not filled out for the manually uploaded file. This field is required to upload extracted data, so it should be captured when you manually register a new document. 

2. When you download the 'register records' and 'document records' csv files, the file name of the manually added document is missing. 

3. Additionally, when you download the 'document files' zip file, the manually added document doesn't appear in the folder.

Definition of Done:
Determine whether the manually uploaded pdfs are being registered/saved in the same way as the csv uploaded documents. 
Update the datagroup downloads ['extracted text csv template', 'document files', 'register records', 'document records'] so that they return necessary fields/files related to manually uploaded documents.

leogi22:
…preserve traceability

Closes #2402 

### Changes:
- data_document.py:  saving filename prior to straight django form save


hausman-gdit (approved):
Approved

PR #2417, opened 2022-10-01
fixing chemical tagset filtering counts

### LINKED ISSUE #2399: Show correct document count on associated tagset filtering on chemical detail page and list presence tag page
This is potentially a bug, but not confirmed - was found while discussing the list presence tag filtering. (LP-tag filtering page has a lot in common with the keyword set filtering on documents in chemical details page).

Visiting https://ccte-factotum.epa.gov/chemical/DTXSID7020637/, the keyword set `{"active_ingredient", "Pesticide"}` has number 30 next to it, suggesting there are 30 documents associated with it. However, when using that filter, the table below shows 34 document entries as opposed to 30.

Acceptance criteria:
- Fix so that the number that shows up next to the set matches exactly the number of associated documents
- On list presence tag page - e.g., https://ccte-factotum-stg.epa.gov/list_presence_tag/502/ - make sure the number of documents shows up in the grey circle

leogi22:
Closes #2399 

### Changes:
- DocumentListJson filterQueryset amended to remove docs that don't exactly match the number of tags present in tagset
- added relevant test: test_associated_list_presence to check clicking on tagset item filters as expected


hausman-gdit (approved):
Approved

PR #2419, opened 2022-10-03
2418-tagset count fix

### LINKED ISSUE #2418: Bug: CP tagset filter missing keyword sets
I was testing the features in dev with the curation team, and we noticed an issue with the CP tagset filter. 
The filter set is correct for most keywords, but we found one where there were less keyword sets than documents. 

https://ccte-factotum-dev.epa.gov/list_presence_tag/493/

The documents table shows 13 documents, but there are only 3 filters that add up to 6 documents. 
When you click on the first filter, which says '2 related documents', it returns 4 documents.

leogi22:
Closes #2418 

### Changes:
- ExtractedListPresence: not filtering tagsets not leading to documents there, rather later in tag_detail.html
- tag_detail.html: not showing tagsets loading to 0 documents

This allows to keep the correct list_presence_tagset forloop number that was clicked on corresponding to allpids list


hausman-gdit (approved):
APproved

PR #2497, opened 2022-12-31
2394 chemical partial search

### LINKED ISSUE #2394: Search for partial keywords on the 'Chemical' search
As a factotum user,
I would like chemical search results to be inclusive of partial keyword matches
So that I can find all related chemicals of interest

Looks like the chemical search is currently only looking for exact matches to search terms. It should be searching through the curated chemical name/CAS and returning partial matches, as well. 

Ex. https://ccte-factotum.epa.gov/search/chemical/?q=cGFyYWJlbg==
When you search for paraben, the chemicals tab only returns two results, whereas we have many different kinds of parabens in Factotum, including but not limited to:
ethylparaben
methylparaben
Butylparaben
Isobutylparaben
Propylparaben
Benzylparaben
Phenylparaben

Definition of Done
Adjust the 'chemicals' tab search to allow for partial keyword matches.

leogi22:
Closes #2394 

### Changes:
- elastic/search_document.py (run_query) changed to limit partial search to chemical tab only
- elastic/search/py changed to extend partial search to tags, functional use and harmonized media


elastic/search_documents.py
```diff
@@ -11,7 +11,7 @@ class SearchDocuments:
     def searchTag(self, term, size, offset=0, page=None):
         search = (
             ListPresenceTag.search()
-            .query("query_string", query=term, fields=["name", "definition"])
+            .query("query_string", query="*" + term + "*", fields=["name", "definition"], analyze_wildcard=True)
```
hausman-gdit, line 14: would this then get an extra set of "*" 's? the query is already wrapped in *'s

leogi22: it doesn't appear to be the case

elastic/search.py
```diff
@@ -240,13 +240,20 @@ def run_query(
...
         else:
-            s = s.query(
-                QueryString(
-                    query = "*" + q + "*",
-                    fields = fields,
-                    analyze_wildcard=True
+            if model == "chemical":
+                s = s.query(
+                    QueryString(
+                        query = "*" + q + "*",
```
hausman-gdit, line 246: doesnt it get passed into here that wraps it already?

leogi22: that code doesn't run for searchtag. searchtag is run inside a 'return'

PR #2502, opened 2023-01-18
2479 adding author and doi to cpcat and lmdocs

### LINKED ISSUE #2479: Add new Fields to CP/LM data
As a factotum admin,
I would like to add more metadata fields for Chemical Presence and Literature Monitoring documents,
so that I can capture and share this information with users. 

Definition of Done:
- Applies to the Chemical Presence and Literature Monitoring Data Types
- Adding new database fields for Author (varchar100) and DOI (varchar50) 
- Update the /datadocument/edit form to display these fields and allow curators to edit. 
  ex. https://ccte-factotum.epa.gov/datadocument/edit/1496320/
- Also need to update the upload registered records csv, to include the new fields upon upload. 
  ex. https://ccte-factotum.epa.gov/datasource/100/datagroup_new/

leogi22:
Closes #2479 

### Changes:
- extractedCPCat and extractedLMDoc model amended
- 2 forms created
- document edit code changed to detect what form should be presented depending on datadocument type

Unable to decide how to change the upload script accordingly (The datadocument upon creation doesn't have yet an extractedText object created and so the two additional fields (author and DOI) should probably be added in another upload)

Needs to run mgration


kbfepa (commented):
Can we add a name to the migration? For readability/searchability and consistency with other migrations

PR #2534, opened 2023-03-10
2494 add search in lp keyword

leogi22:
Closes #2494 add search in lp keyword

### Changes:
- tag_detail to remove download buttons and embed them in the table design
- lp_tag_tables.js to enable searching on 4 out of 5 fields to use django search functionality
- ajax.py to correctly name the chemical columns and enable searching on chemical page columns, and allow searching on document page by using django built-in feature on model field document_title and combining it with ad-hoc searching on tagset column (that search is faster)


kbfepa (approved):
The data document name search doesn't behave like the other pages' data document name search, but merging it in for now. Searching on tagset filtering seems tricky, so nicely done, thanks.

PR #2539, opened 2023-03-16
2492 align nav items 

hausman-gdit:
Closes FACTOTUM-2397

I fiddled with a fair amount of in-betweens, icons, spacing, sizing. 

### Changes:
- removed item borders
- set inline


kbfepa (approved):
This is great, this also fixes the weird spacing that shows up on the very end when a user is logged in - clean!

PR #2510, opened 2023-02-03
2481 add fields for lm docs

leogi22:
Closes #2481

### Changes:
- stat_note added to statistical value model and extraction scripts
- fields amendment for LMrec and LMDoc
- test fixes

PR #2554, opened 2023-04-17
2510 chemical card formatting

hausman-gdit:
Closes #2510

This really works best at a "full size" screen, when using a laptop the chemical cards *should* just move down.

### Changes:
- decreased `font-size` from `150%` to `110%`
- swapped out `h3` w/ `h4`
- tweaked the column sizing
- on hover of curated chem name now also shows the chemical name (can see the name even when truncated)


kbfepa (approved):
Side by side comparisons (same size/zoom) shows more of the card is clearly showing. (Might be an easy way to demo on a large monitor).
It also is aesthetically a lot more pleasing!
SID still can get cut off when page is thinner, but more of it shows, and the fact that ChemExpo won't have the side bar to the right may make this even better there - TBD.

